PR #52, opened 2021-09-28
Level solver counterexample construction

ice1000:
This PR aims at testing bors.


ice1000:
> Note: that bors reads bors.toml from the pull requests it’s merging, not the one in master, so changes to the file get checked before they land. You can therefore verify whether bors is working by using bors r+ to try and land the PR that enables it!

https://bors.tech/documentation/getting-started/

🤦‍♀️

PR #74, opened 2021-10-09
Relicense as MIT

ice1000:
I marked the coauthorship based on their participation in the internal discussion on relicensing. I did not update all files because we may automatically change them while committing in IntelliJ, and I don't want to make a huge commit.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Yinsen (Tesla) Zhang
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


re-xyr:
> so if it's displayed as plain text, it will be a huge horizontal line. 

I think you should wrap it at 80 chars.

ice1000:
Anyways, I copied the GitHub MIT license template and it's wrapped at 80 chars, so it should be fine now.

re-xyr:
> Is it standard practice?

That is what I’ve seen in most license files. You can see the ones GitHub generates for new repos.

ice1000:
> > Is it standard practice?
> 
> That is what I’ve seen in most license files. You can see the ones GitHub generates for new repos.

Yes, that's where I get the updated version. Does it look good to you now?

re-xyr:
Yes.

Issue #29, opened 2021-09-20
'Success' test cases deduplication


ice1000:
Currently, we have lots of code duplicated in the 'success' test folder. How much of them are actually duplicated, and how much can we delete them? I think a lot of files can be removed.

I recently added an extremely heavy sanity check to the tycker, and it's causing some OOM on my laptop when running tests. These checks are removed in release builds, but we still want the tests to run faster.

PR #79, opened 2021-10-11
Run custom test in TestRunner

imkiva:
Seen #29 
> Another problem is we can’t run a certain success/failure case (at least not that I know of). We can only run all of them in a batch which doesn’t seem reasonable in every scenario.

 Gradle does not support passing arguments from the command line to JUnit tests. So a separate `main()` method was used to make this possible.

single test file or directory containing test files are accepted as valid test target

Usage:
image

from command line:
```shell
gradle runCustomTest --args <file1> <file2> ...
```

Any ideas on making this more useful?

### DIFF
--- a/base/build.gradle.kts
+++ b/base/build.gradle.kts
@@ -37,3 +37,9 @@ tasks.named<Test>("test") {
   testLogging.showCauses = true
   inputs.dir(projectDir.resolve("src/test/resources"))
 }
+
+tasks.register<JavaExec>("runCustomTest") {
+  group = "Execition"
+  classpath = sourceSets.test.get().runtimeClasspath
+  main = "org.aya.test.TestRunner"
+}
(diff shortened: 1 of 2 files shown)


ice1000:
> Usage:

Can you use command line to invoke it? Like using some gradle hacks?

imkiva:
> > Usage:
> 
> Can you use command line to invoke it? Like using some gradle hacks?

Not found yet. I want command line invocation too

ice1000:
Relevant:

+ https://stackoverflow.com/q/36843964/7083401
+ https://stackoverflow.com/a/69384805/7083401 (LMAO, they even used the same class name as ours)

ice1000:
Just change it to `sourceSets.test`

ice1000:
Dunno if it would work tho

imkiva:
> Relevant:
> 
> * https://stackoverflow.com/q/36843964/7083401
> * https://stackoverflow.com/a/69384805/7083401 (LMAO, they even used the same class name as ours)

Yeah, but we need passing arguments to it. Gradle needs some hack like:

- https://stackoverflow.com/a/42492844

PR #85, opened 2021-10-16
Repl prototype

ice1000:
This code introduces a new dependency and is very inconvenient to switch branches back and forth (Gradle script changes when you switch branch, and you need to rebuild your project each time you switch branch). So I decided to merge the code at this time

PR #86, opened 2021-10-16
Some follow-up changes in comments

ice1000:
This commit addresses some comments in #85 by me and @imkiva.

### DIFF
--- a/cli/src/main/java/org/aya/cli/repl/AbstractRepl.java
+++ b/cli/src/main/java/org/aya/cli/repl/AbstractRepl.java
@@ -20,8 +20,7 @@ public abstract class AbstractRepl implements Closeable {
   public static final @NotNull @Nls String APP_NAME = "Aya REPL";
   public static final @NotNull @Nls String INTRODUCTION_MESSAGE = APP_NAME + " " + GeneratedVersion.VERSION_STRING;
 
-  // TODO: what locator and builder?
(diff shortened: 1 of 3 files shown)


ice1000:
Here's a summary of what's missing from the current impl:

+ [ ] Lots of methods and fields are package-private, and I don't know why. I guess making more of them `private` is a better idea.
+ [ ] Parser is unimplemented.
+ [ ] Completion is unimplemented. I added completion for keywords, but that's only available for non-commands. If your line starts with `:`, then no completion. This should be changed after we have commands that accept expressions.
+ [ ] No repl commands. I think we need a general interface for repl commands so the repl framework is extensible. Here's a list of what I'm looking for.
  + [ ] `:t`, `:type` from GHCi. Completion should be enabled for this command.
  + [ ] `:normalize` or `:eval` -- whatever you'd like to call it -- for normalizing the expression. Completion should be enabled.
  + [ ] `:prompt` that changes the prompt.
  + [ ] An option to change the pretty printing width.
  + [ ] An option to change pretty printing options. Available options should be completed.
+ [ ] Save history for jline3 based repl. We may start using `~/.aya/something` for this purpose.
+ [ ] Handle redefinitions.
+ [ ] Completion of commands. This is very important.

ice1000:
Everything in the above list are implemented in the Arend repl, so maybe we can do some copy-paste.

Issue #87, opened 2021-10-16
Improve the REPL


ice1000:
Here's a summary of what's missing from the current impl:

+ [ ] Lots of methods and fields are package-private, and I don't know why. I guess making more of them `private` is a better idea.
+ [ ] Parser is unimplemented.
+ [ ] Completion is unimplemented. I added completion for keywords, but that's only available for non-commands. If your line starts with `:`, then no completion. This should be changed after we have commands that accept expressions.
+ [ ] No repl commands. I think we need a general interface for repl commands so the repl framework is extensible. Here's a list of what I'm looking for.
  + [ ] `:t`, `:type` from GHCi. Completion should be enabled for this command.
  + [ ] `:normalize` or `:eval` -- whatever you'd like to call it -- for normalizing the expression. Completion should be enabled.
  + [ ] `:prompt` that changes the prompt.
  + [ ] An option to change the pretty printing width.
  + [ ] An option to change pretty printing options. Available options should be completed.
+ [ ] Save history for jline3 based repl. We may start using `~/.aya/something` for this purpose.
+ [ ] Handle redefinitions.
+ [ ] Completion of commands. This is very important.

_Originally posted by @ice1000 in https://github.com/aya-prover/aya-dev/issues/86#issuecomment-944882637_

ice1000:
@refparo How would you expect, say, to change a code using `~/.aya/something` to a code using XGD base dir? I looked up the link and I expected 'read $XDG_STATE_HOME, and if it's null, use `~/.aya`, am I right?

refparo:
> If `$XDG_STATE_HOME` is either not set or empty, a default equal to `$HOME/.local/state` should be used.

According to [the specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

ice1000:
@refparo Are you sure that everyone wants to use `$HOME/.local/state/something`? What if two different programs both use, say, `$HOME/.local/state/history`?

ice1000:
Or should I create a subdir under `$HOME/.local/state`, like `$HOME/.local/state/aya`?

refparo:
Yeah, use `$HOME/.local/state/aya_history`, or `$HOME/.local/state/aya/history`. It's just like how people use `$HOME/.config` (which is the `$XDG_CONFIG_HOME`)

ice1000:
Great

ice1000:
[...]

PR #91, opened 2021-10-16
Some repl enhancements

ice1000:
image

Right now I'm using the hack that @ShreckYe uses to run commands, but it should be changed to an extensible framework later.

Complete two subtasks of #87.


ice1000:
Empty input crashes the repl. I think this shouldn't happen @ShreckYe

ice1000:
bors merge

ShreckYe:
> Empty input crashes the repl. I think this shouldn't happen @ShreckYe

Is this fixed by this PR?

ice1000:
@ShreckYe Yes

PR #100, opened 2021-10-20
Pattern domination warning, disable confluence check for incorrect patterns

### LINKED ISSUE #99: Clause "redundancy" or "domination" warning
Thanks for detailed explanations from you both. If "true" match all the cases, maybe also helpful to users that the other clause is "unreachable" because of it. Just tried this similar expression in Arend:
```
\data Bool | real | false
\func not (b : Bool) : Bool
 | true => false
 | false => real
```
It gives warning below:
```
[WARNING] t:4:4: Clause is redundant
  In: | false => real
```

_Originally posted by @nobodxbodon in https://github.com/aya-prover/aya-dev/discussions/96#discussioncomment-1500361_

ice1000:
Fixes #99 

This PR is #100! Welcome to the three-digits-issue-number era!


ice1000:
I think there's a bug.

ice1000:
Ok, 14 seconds ahead of me.

PR #108, opened 2021-10-21
ERROR: improve error report (#96)

imkiva:
- Add line break between each problem
- Count errors and warnings

Signed-off-by: imkiva <imkiva@islovely.icu>

### DIFF
--- a/api/src/main/java/org/aya/api/error/CountingReporter.java
+++ b/api/src/main/java/org/aya/api/error/CountingReporter.java
@@ -2,26 +2,38 @@
 // Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
 package org.aya.api.error;
 
+import kala.collection.mutable.MutableMap;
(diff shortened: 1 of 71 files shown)


ice1000:
Also: maybe we should find a better way to maintain the test cases. Right now it's tracking all the details of the compiler output, and maybe we don't wanna do that.

imkiva:
Does this mean _this catch clause is unreachable_?

image

imkiva:
> Sometimes these are just inaccurate. Try setting a break point if you are really curious.

Breakpoint in `FileModuleLoader` does not hit after all tests are passed. I think we can remove these catch clauses because `SingleFileCompiler` also caught these exceptions whose breakpoint was hit.

ice1000:
> btw, I decide to squash this PR since this feature is so simple

/cc https://github.com/bors-ng/bors-ng/issues/1028

PR #131, opened 2021-10-30
New REPL command system and load/cd/pwd commands

imkiva:
This PR:
- Implemented type-based completer detection and conversation which support toggling.
- Added new `:cd`, `:load`, `:pwd` commands using the new command system.

### How to add a new completer associated with a type:
https://github.com/aya-prover/aya-dev/blob/0ed3e46076bd829ebd4f62df734dc15a117425af/cli/src/main/java/org/aya/cli/repl/Repl.java#L49-L56

  The line
  ```java
  CommandArg.from(Path.class, new Completers.FileNameCompleter(), this::resolveFile)
  ```
  tells the command manager that `Path` is a valid command argument, and it should be completed using file name completer and converted from `String` using `this::resolveFile`

For more usage, see [ReplCommands.java](https://github.com/aya-prover/aya-dev/blob/0ed3e46076bd829ebd4f62df734dc15a117425af/cli/src/main/java/org/aya/cli/repl/ReplCommands.java)


cli/src/main/java/org/aya/cli/repl/ReplCommands.java
```diff
@@ -0,0 +1,111 @@
...
+import org.aya.cli.repl.command.Command;
+import org.aya.pretty.doc.Doc;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public interface ReplCommands {
```
ice1000, line 16: Do you think it makes sense to make a constructor for Command that takes a lambda?

ice1000 (commented):
tql

cli/src/main/java/org/aya/cli/repl/ReplCommands.java
imkiva: It is impossible. The `@Entry` method can take arbitrary arguments and we will provide completion by checking their types. Lambdas can only have fixed arguments? Are there any container that supports storing vararg (not `Object`) lambdas?

cli/src/main/java/org/aya/cli/repl/Repl.java
```diff
@@ -76,32 +110,33 @@ private boolean singleLoop() {
...
-      } else printResult(evalWithContext(line));
+      } else printResult(eval(line));
     } catch (EOFException ignored) {
       // user send ctrl-d
       return false;
     } catch (InterruptedException ignored) {
       // user send ctrl-c
       return true;
-    } catch (Exception e) {
+    } catch (Throwable e) {
```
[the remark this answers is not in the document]
imkiva: We used `MethodHandle#invoke` which throws `Throwable` instead of `Exception`

cli/src/main/java/org/aya/cli/repl/ReplCommands.java
ice1000: I see, so it's based on annotation processing!!

imkiva: Yeah!

ice1000:
Completion for command name after `:help` is lost. Can you add it back?

imkiva:
> Completion for command name after `:help` is lost. Can you add it back?

see https://github.com/aya-prover/aya-dev/pull/131/commits/64879e9524b39f2df61298f826b17986a6f6974e

It is so easy to add completion for a command!

ice1000:
> > Completion for command name after `:help` is lost. Can you add it back?
> 
> see [64879e9](https://github.com/aya-prover/aya-dev/commit/64879e9524b39f2df61298f826b17986a6f6974e)
> 
> It is so easy to add completion for a command!

Perfect. Here's one more random idea: stop using multiple names for commands, but instead when we see a command, we browse all commands and see if the given string is a "beginning substring" of some commands. If there's one fit, then use that command (say, `:q`, `:qu`, `:qui` will all refer to `:quit`, and `:h`, `:he` will refer to `:help`, but :`p` causes ambiguity, where one must instead write `:pw` or `:pp`).

Do you think it's a good idea?

bors merge

imkiva:
> Perfect. Here's one more random idea: stop using multiple names for commands, but instead when we see a command, we browse all commands and see if the given string is a "beginning substring" of some commands. If there's one fit, then use that command (say, `:q`, `:qu`, `:qui` will all refer to `:quit`, and `:h`, `:he` will refer to `:help`, but :`p` causes ambiguity, where one must instead write `:pw` or `:pp`).

I love this idea! Coming in new PR.

ice1000:
> > Perfect. Here's one more random idea: stop using multiple names for commands, but instead when we see a command, we browse all commands and see if the given string is a "beginning substring" of some commands. If there's one fit, then use that command (say, `:q`, `:qu`, `:qui` will all refer to `:quit`, and `:h`, `:he` will refer to `:help`, but :`p` causes ambiguity, where one must instead write `:pw` or `:pp`).
> 
> I love this idea! Coming in new PR.

Rationale:

image

Isn't this confusing? :rofl:

ice1000:
Also

```
D:\git-repos\aya-prover [ice ×]> ghci
GHCi, version 9.0.1: https://www.haskell.org/ghc/  :? for help
ghci> :qui
Leaving GHCi.
```

So :D

ice1000:
The only exception is that `?` might be a useful alias of `:help`, but right now we're not parsing `:?` as a token

imkiva:
> > > Perfect. Here's one more random idea: stop using multiple names for commands, but instead when we see a command, we browse all commands and see if the given string is a "beginning substring" of some commands. If there's one fit, then use that command (say, `:q`, `:qu`, `:qui` will all refer to `:quit`, and `:h`, `:he` will refer to `:help`, but :`p` causes ambiguity, where one must instead write `:pw` or `:pp`).
> > 
> > 
> > I love this idea! Coming in new PR.
> 
> Rationale:
> 
> image
> 
> Isn't this confusing? 🤣

_Higher order command_

Not confusing actually. because users may want to know how to use `:help` command.

image

ice1000:
Nah, I'm talking 'bout that there are multiple names like `l` and h`, these names are quite confusing

imkiva:
> Nah, I'm talking 'bout that there are multiple names like `l` and `h`, these names are quite confusing

Well yes. I agree.

imkiva:
> The only exception is that `?` might be a useful alias of `:help`, but right now we're not parsing `:?` as a token

I changed the parser behavior in completion mode. It now parses line using a shell-like way. So `:?` will be parsed as a token I guess.

ice1000:
Wait what😂 Sorry for my careless review. I didn't even notice it.

`:t (Ty` completes nothing. I think it's a bad idea.

imkiva:
> Wait what😂 Sorry for my careless review. I didn't even notice it.
> 
> `:t (Ty` completes nothing. I think it's a bad idea.

Oh. I think this can be made type-directed too....?

ice1000:
`(Ty` itself also completes nothing. I think the very point of the parser is to work with completion.

ice1000:
> Oh. I think this can be made type-directed too....?

Wdym by type-directed? I think the reason why you want shell-like parsing is that you want `"hey ya"` to be a token. Since Aya lexer supports strings already, I think it's okay to use the Aya lexer as-is.

imkiva:
> > Oh. I think this can be made type-directed too....?
> 
> Wdym by type-directed? I think the reason why you want shell-like parsing is that you want `"hey ya"` to be a token. Since Aya lexer supports strings already, I think it's okay to use the Aya lexer as-is.

I want `../abc` to be a token instead of `Tokens[.., /, abc]`

imkiva:
> Wdym by type-directed? 

Like, for `Path` argument we use shell-like parser.
for `Code` argument we use aya parser

ice1000:
> > Wdym by type-directed?
> 
> Like, for `Path` argument we use shell-like parser. for `Code` argument we use aya parser

I see. Are you sure that `DefaultParser` is a shell-like parser?

imkiva:
> > > Wdym by type-directed?
> > 
> > 
> > Like, for `Path` argument we use shell-like parser. for `Code` argument we use aya parser
> 
> I see. Are you sure that `DefaultParser` is a shell-like parser?

Yes, by reading its code throughly.

ice1000:
> I want `../abc` to be a token instead of `Tokens[.., /, abc]`

How hard it would be, to change the file name completer to work with `Tokens[.., /, abc]` instead of just one token?

imkiva:
> I see. Are you sure that DefaultParser is a shell-like parser?

Proof:
> `(Ty` itself also completes nothing. I think the very point of the parser is to work with completion.

use `( <space> Ty <tab>` will give completions.

ice1000:
> > I see. Are you sure that DefaultParser is a shell-like parser?
> 
> Proof:
> 
> > `(Ty` itself also completes nothing. I think the very point of the parser is to work with completion.
> 
> use `( <space> Ty <tab>` will give completions.

Yes. I used `AyaReplParser` *only because* I want `(Ty` to be completed with `(Type`. See https://github.com/JetBrains/Arend/pull/312#issuecomment-951517556

ice1000:
And now it's gone so I'm very sad😢

imkiva:
> > I want `../abc` to be a token instead of `Tokens[.., /, abc]`
> 
> How hard it would be, to change the file name completer to work with `Tokens[.., /, abc]` instead of just one token?

I don't try. But my cloud programmer mind tells me that we need to traverse the word list and see if a word is a valid path item and joining them finally. This is platform-specific and too mind-consuming.

ice1000:
> Question: will aya parser keep spaces in parsed words? If so, then it is easy.

I don't get the question. Care to exemplify?

imkiva:
> > Question: will aya parser keep spaces in parsed words? If so, then it is easy.
> 
> I don't get the question. Care to exemplify?

Like `:cmd  Nat  git/Nat.aya`

If spaces remain in tokens, we have `Tokens["Nat", " ", "git", "/", "Nat", ".", "aya"]`. We only traverse the word list from `indexOf(" ")` to `wordIndex`, and then we have a Path.

if not, we have `Tokens["Nat", "git", "/", "Nat", ".", "aya"]`. We cannot tell which parts is the tokens of the second argument.

ice1000:
Just search for `filter` in the completer. I think a smart person like you should get it immediately

imkiva:
But I still recommend make parser type-directed. It reuses the code and reduces conditions.

ice1000:
> But I still recommend make parser type-directed. It reuses the code and reduces conditions.

It's very hard. It's gonna hardcode some rules

imkiva:
> > But I still recommend make parser type-directed. It reuses the code and reduces conditions.
> 
> It's very hard. It's gonna hardcode some rules

I think it is just like what we did in `CmdCompleter`

ice1000:
I really want to blame the file name completer! It does not seem to me that multiple tokens would break it

ice1000:
I really believe that testing the REPL is essential. I need to test completion as well😭 at least it should remind people to not break implemented features.

imkiva:
> I really believe that testing the REPL is essential. I need to test completion as well😭 at least it should remind people to not break implemented features.

Then let's do some copy-paste-modify work?

ice1000:
> > I really want to blame the file name completer! It does not seem to me that multiple tokens would break it
> 
> Then let's do some copy-paste-modify work?
> 
> Edit: fixed wrong quotion

I just don't know how hard it would be. I think this is a difficult problem in general.

ice1000:
If it is very very hard, I don't think it is worth implementing. I just don't want to have it overcomplicated

imkiva:
> If it is very very hard, I don't think it is worth implementing. I just don't want to have it overcomplicated

Yes. I read its code and realized it is too hard to maintain it on our own. Like, it needs to parse environmental variable `LS_COLOR` to make candidates colorized. It is mainly about string operation, which is too tiring qwq

ice1000:
@imkiva What about completion then? Currently `AyaReplParser` is useless, and is essentially equivalent to using `DefaultParser`

imkiva:
> What about completion then? 

It's quite common (parsing the `~` to user's home). You can find its source here: https://github.com/jline/jline3/blob/997496e6a6338ca5d82c7dec26f32cf089dd2838/builtins/src/main/java/org/jline/builtins/Completers.java#L331

> Currently `AyaReplParser` is useless, and is essentially equivalent to using `DefaultParser`

`AyaReplParser` is also used in highlighter.

ice1000:
> It's quite common (parsing the ~ to user's home).

I see. What do you think about having our own version? Java Path works with `/` on Windows and I don't really think coloring the directories is important, so a lot of things can be removed.

> `AyaReplParser` is also used in highlighter.

I see

imkiva:
@ice1000 I think coloring is very important for me. I am so used to see my colorful terminal.

PR #133, opened 2021-11-01
REPL: choose parser by param type (currently only Path.class uses a shell like parser)

imkiva:
This adds back completion for `(Ty <tab>` and also supports path completion.

### DIFF
--- a/cli/src/main/java/org/aya/cli/repl/jline/AyaReplParser.java
+++ b/cli/src/main/java/org/aya/cli/repl/jline/AyaReplParser.java
@@ -16,7 +16,11 @@ import org.jline.reader.impl.DefaultParser;
 import java.util.Collections;
 import java.util.List;
 
-public record AyaReplParser(@NotNull CommandManager cmd) implements Parser {
+public record AyaReplParser(@NotNull CommandManager cmd, @NotNull DefaultParser shellLike) implements Parser {
+  public AyaReplParser(@NotNull CommandManager cmd) {
+    this(cmd, new DefaultParser());
+  }
+
   public record AyaParsedLine(
     int wordCursor,
     @NotNull List<@NotNull String> words,
@@ -70,9 +74,8 @@ public record AyaReplParser(@NotNull CommandManager cmd) implements Parser {
     if (trim.startsWith(Command.PREFIX)) {
       var shellLike = cmd.parse(trim.substring(1)).command()
         .flatMap(c -> Option.of(c.argFactory()))
-        .map(CommandArg::shellLike)
-        .getOrDefault(false);
-      if (shellLike) return new DefaultParser().parse(line, cursor, context);
+        .map(CommandArg::shellLike).getOrDefault(false);
+      if (shellLike) return shellLike().parse(line, cursor, context);
     }
     // Drop whitespaces
     var tokens = tokensNoEOF(line)
--- a/cli/src/main/java/org/aya/cli/repl/jline/AyaReplParser.java
+++ b/cli/src/main/java/org/aya/cli/repl/jline/AyaReplParser.java
@@ -3,16 +3,20 @@
 package org.aya.cli.repl.jline;
 
 import kala.collection.SeqView;
+import kala.control.Option;
 import org.antlr.v4.runtime.Token;
 import org.aya.cli.repl.command.Command;
+import org.aya.cli.repl.command.CommandArg;
+import org.aya.cli.repl.command.CommandManager;
 import org.aya.concrete.parse.AyaParsing;
 import org.jetbrains.annotations.NotNull;
 import org.jline.reader.*;
+import org.jline.reader.impl.DefaultParser;
 
 import java.util.Collections;
 import java.util.List;
 
-public class AyaReplParser implements Parser {
+public record AyaReplParser(@NotNull CommandManager cmd) implements Parser {
   public record AyaParsedLine(
     int wordCursor,
     @NotNull List<@NotNull String> words,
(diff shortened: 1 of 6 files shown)


imkiva:
CI failed. But I am going to fix it tomorrow. Good night ~

ice1000:
@imkiva Nighty night!

cli/src/main/java/org/aya/cli/repl/jline/AyaReplParser.java
```diff
@@ -63,8 +66,14 @@ private ParsedLine debug(String line, int cursor, ImmutableSeq<Token> tokens, Ay
...
     }
-    // use a shell like parser for completion only
-    if (context == ParseContext.COMPLETE) return new DefaultParser().parse(line, cursor, context);
+    var trim = line.trim();
+    if (trim.startsWith(Command.PREFIX)) {
+      var shellLike = cmd.parse(trim.substring(1)).command()
+        .flatMap(c -> Option.of(c.argFactory()))
+        .map(CommandArg::shellLike)
+        .getOrDefault(false);
+      if (shellLike) return new DefaultParser().parse(line, cursor, context);
```
ice1000, line 75: Should maintain an instance instead of instantiating it every single time

cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java
```diff
@@ -49,14 +51,23 @@ public Context(@NotNull Repl repl) {
       this.repl = repl;
     }
 
+    private @NotNull Tuple2<String, Boolean> fixWord(@NotNull String word, ParsedLine line) {
+      if (word.startsWith(":") || word.startsWith(Constants.SCOPE_SEPARATOR)) {
```
ice1000, line 55: COMMAND_PREFIX

cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java
```diff
@@ -49,14 +51,23 @@ public Context(@NotNull Repl repl) {
...
       var word = line.word();
+      var fixed = fixWord(word, line);
       var context = repl.replCompiler.getContext();
       context.modules.view().forEach((mod, contents) -> {
         var modName = mod.joinToString(Constants.SCOPE_SEPARATOR) + Constants.SCOPE_SEPARATOR;
-        if (!modName.startsWith(word)) return;
+        if (!modName.startsWith(fixed._1)) return;
         contents.keysView()
-          .map(name -> modName + name)
+          .map(name -> fixed._2 ? Constants.SCOPE_SEPARATOR + name : modName + name)
```
ice1000, line 70: `( ? : ) + name` should be better, agree?

ice1000:
Did you adapt the new command parsing strategy¿

ice1000 (approved):
Overall, I think this hack is fine. Well done.

cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java
imkiva: This `:` is not a command prefix but a part of scope separator 🤣

cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java
imkiva: Yes!

imkiva:
> Did you adapt the new command parsing strategy¿

what is that?

imkiva:
In case we need a homemade completer some day 🤣 
```patch
Index: cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java b/cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java
--- a/cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java	(revision 7a89b496df2ee4a599182c7d06bed4f9104b34c1)
+++ b/cli/src/main/java/org/aya/cli/repl/jline/completer/AyaCompleters.java	(revision b41784a6a4f0461289e742b579f30ece69ff86f9)
@@ -2,6 +2,7 @@
 // Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
 package org.aya.cli.repl.jline.completer;
 
+import kala.collection.Seq;
 import kala.tuple.Tuple;
 import kala.tuple.Tuple2;
 import org.aya.cli.repl.Repl;
@@ -15,6 +16,9 @@
 import org.jline.reader.LineReader;
 import org.jline.reader.ParsedLine;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.List;
 
@@ -76,6 +80,44 @@
     }
   }
 
+  record File(@NotNull Repl repl) implements Completer {
+    @Override public void complete(LineReader reader, ParsedLine line, List<Candidate> candidates) {
+      int wordIndex = line.wordIndex();
+      // find words between [the space nearest to wordIndex, wordIndex]
+      var startOpt = Seq.from(line.words()).withIndex()
+        .filter(i -> i._2.matches("\\s+"))
+        .filter(i -> i._1 < wordIndex)
+        .map(i -> i._1).lastOption();
+      if (startOpt.isEmpty() || startOpt.get() >= wordIndex) return;
+      int start = startOpt.get();
+      var wordBeforeCursor = line.word().substring(0, line.wordCursor());
+      var argument = String.join("", line.words().subList(start + 1, wordIndex)) + wordBeforeCursor;
+      var path = repl.resolveFile(argument);
+      var isDir = Files.isDirectory(path);
+      try {
+        Files.list(isDir ? path : path.getParent()).filter(this::accept).forEach(f -> {
+          var text = f.getFileName().toString();
+          var value = wordBeforeCursor.substring(0, wordBeforeCursor.lastIndexOf(java.io.File.separator) + 1) + text;
+          if (Files.isDirectory(f)) candidates.add(new Candidate(
+            value + "/", text + "/", null, null, "/", null, false
+          ));
+          else candidates.add(new Candidate(
+            value, text, null, null, null, null, true));
+        });
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
+    }
+
+    private boolean accept(@NotNull Path path) {
+      try {
+        return !Files.isHidden(path);
+      } catch (IOException e) {
+        return false;
+      }
+    }
+  }
+
   record Help(@NotNull Repl repl) implements Completer {
     @Override public void complete(LineReader reader, ParsedLine line, List<Candidate> candidates) {
       repl.commandManager.cmd.view().map(CommandManager.CommandGen::owner)

```

ice1000:
> > Did you adapt the new command parsing strategy¿
> 
> what is that?

To choose a command if the typed command is the prefix of a unique command instead of providing shorthands of command names

imkiva:
> > > Did you adapt the new command parsing strategy¿
> > 
> > 
> > what is that?
> 
> To choose a command if the typed command is the prefix of a unique command instead of providing shorthands of command names

I see. it is not implemented yet. ( It is in my local branch 😃 )

PR #138, opened 2021-11-04
Allow emission of univ, desugar to Type0

ice1000:
bors r+


ice1000:
> Build succeeded:
> 
> * [gradle-check](https://github.com/aya-prover/aya-dev/runs/4109177079?check_suite_focus=true)

Thanks!

PR #145, opened 2021-11-06
Tyck of one SCC will not be interrupted by unrelated failed SCCs

imkiva:
In the past, we will stop tycking immediately when a statement fails. Now we can skip the failed statement and its dependents and try to tyck as much as possible and always produce a `wellTyped` list, which is quite good for our LSP frontend.

### DIFF
--- a/base/src/main/java/org/aya/tyck/order/IncrementalTycker.java
+++ b/base/src/main/java/org/aya/tyck/order/IncrementalTycker.java
@@ -1,0 +1,45 @@
+// Copyright (c) 2020-2021 Yinsen (Tesla) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.tyck.order;
+
+import kala.collection.SeqLike;
+import kala.collection.mutable.MutableSet;
+import org.aya.concrete.resolve.ResolveInfo;
+import org.aya.concrete.stmt.Decl;
+import org.aya.concrete.stmt.Stmt;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Incremental and non-stopping compiler for SCCs.
+ *
+ * @author kiva
+ */
+public record IncrementalTycker(
+  @NotNull SCCTycker sccTycker,
+  @NotNull ResolveInfo resolveInfo,
+  @NotNull MutableSet<Stmt> skipped
+) {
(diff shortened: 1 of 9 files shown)


base/src/main/java/org/aya/tyck/order/IncrementalTycker.java
```diff
@@ -0,0 +1,43 @@
...
+  @NotNull MutableSet<Stmt> skipped
+) {
+  public IncrementalTycker(@NotNull SCCTycker sccTycker, @NotNull ResolveInfo resolveInfo) {
+    this(sccTycker, resolveInfo.toUsageInfo(), MutableSet.of());
+  }
+
+  public void tyckSCC(@NotNull ImmutableSeq<Stmt> scc) {
+    try {
+      // we are more likely to check correct programs.
+      // I'm not sure whether it's necessary to optimize on our own.
```
ice1000, line 29: To optimize what?

lsp/src/main/java/org/aya/lsp/server/AyaService.java
```diff
@@ -63,7 +63,7 @@ public void registerLibrary(@NotNull Path path) {
     var compilerFlags = new CompilerFlags(
       CompilerFlags.Message.EMOJI, false, null,
       libraryManager.modulePath.view());
-
+    libraryManager.loadedFiles.remove(filePath);
```
ice1000, line 66: What does this do?

tools/src/main/java/org/aya/util/MutableGraph.java
```diff
@@ -39,6 +39,12 @@ private boolean hasPath(@NotNull MutableSet<T> book, @NotNull T from, @NotNull T
     return new Tarjan().tarjan();
   }
 
+  public @NotNull MutableGraph<T> transpose() {
+    var tr = MutableGraph.<T>empty();
```
ice1000, line 43: I propose rename to `create ()`

lsp/src/main/java/org/aya/lsp/server/AyaService.java
imkiva: Remove last time compiled core. Fix vscode re-highlighting (need to change frontend code, too)

base/src/main/java/org/aya/tyck/order/IncrementalTycker.java
imkiva: See the next two lines of this comment.

ice1000:
Cool!!

PR #146, opened 2021-11-06
KALA: upgrade to 0.30.0

Glavo:


base/src/main/java/org/aya/core/visitor/VarConsumer.java
```diff
@@ -113,7 +113,9 @@
         bound.append(param.ref());
         param.type().accept(this, Unit.unit());
       });
-      bound.removeAt(start, term.params().size());
+      for (int i = 0; i < term.params().size(); i++) {
+        bound.removeAt(start); // TODO: DynamicSeq::removeAt(int, int)
```
ice1000, line 117: Can you update Kala with this method?

ice1000 (commented):
Can you also replace the LinkedHashMaps with the ones in Kala?

base/src/main/java/org/aya/core/visitor/VarConsumer.java
Glavo: > Can you update Kala with this method?

There are usually two styles of APIs that need to specify a range:

1. Accept an index and a length;
2. Accept begin index and end index.

The original `removeAt(int, int)` is inherited from Scala's `remove`, which belongs to the first style. However, Kala's overall API design tends to the second style, which makes it a little out of place.

For this reason, I have temporarily removed this API. I need to spend some time exploring it again and reintroducing it in a new form in the future.

Glavo:
> Can you also replace the LinkedHashMaps with the ones in Kala?

Frankly, I am not satisfied with the current implementation of HashMap. 

Implementing `LnkedHashMap` requires reusing most of the implementation of `HashMap`. I hope to do it after re implementing `HashMap`.

ice1000:
I see, thanks!

bors merge

PR #165, opened 2021-11-11
Cherry pick from new bind block

imkiva:


api/src/main/java/org/aya/api/distill/AyaDocile.java
```diff
@@ -10,8 +10,14 @@
  * @author ice1000
  */
 @FunctionalInterface
-@Debug.Renderer(text = "toDoc(DistillerOptions.DEBUG).debugRender()")
+@Debug.Renderer(text = "debuggerOnlyToDoc().debugRender()")
 public interface AyaDocile /*extends Docile*/ {
+  // load DistillerOptions by using it explicitly so IDEA won't show cannot load blahblah
+  // in the debugger window. this should not be used in any other place.
```
ice1000, line 16: I propose making it javadoc

base/src/main/java/org/aya/concrete/stmt/Decl.java
```diff
@@ -269,12 +271,17 @@ public StructField(
...
       this.body = body;
+      this.operator = operator;
       this.ref = DefVar.concrete(this, name);
     }
 
     @Override public @NotNull DefVar<? extends Def, StructField> ref() {
       return ref;
     }
+
+    @Override public @Nullable Operator asOperator() {
```
ice1000, line 282: Actually, I believe `getOperator` is a better name.

ice1000: Can you rename the base interface as well?

ice1000 (commented):
```java
  public static final class DataDecl extends Decl implements OpDecl {
    public final @NotNull DefVar<DataDef, DataDecl> ref;
    public @NotNull Expr result;
    public final @NotNull ImmutableSeq<DataCtor> body;
    public @Nullable Operator operator;
```

Do you agree that the above `operator` can be made final?

ice1000 (approved):
bors merge

what else could I say?

PR #166, opened 2021-11-11
PP options

### LINKED ISSUE #39: How do we load default PP config?
From CLI args? From a file? Wdyt?

### LINKED ISSUE #120: Further improve the REPL
+ [x] An option to change pretty printing options. Available options should be completed.
+ [x] Module redefinitions.
+ [x] Completion for variables in the current scope.
+ [x] A command for loading an Aya file into the repl.
+ [x] Test the repl.

Feel free to pick and implement these

ice1000:
+ Refactor: make `DistillerOptions` a map from a `enum Key` to `boolean`
+ Fix #120: implement `:print-options` command in the obvious way
+ Fix #39: load default pp config from the repl's config

bors d=imkiva

### DIFF
--- a/base/src/main/java/org/aya/distill/ConcreteDistiller.java
+++ b/base/src/main/java/org/aya/distill/ConcreteDistiller.java
@@ -50,7 +50,7 @@ public class ConcreteDistiller extends BaseDistiller implements
   }
 
   @Override public Doc visitLam(Expr.@NotNull LamExpr expr, Outer outer) {
-    if (!options.showImplicitPats && !expr.param().explicit()) {
+    if (!(boolean) options.map.get(DistillerOptions.Key.ShowImplicitPats) && !expr.param().explicit()) {
       return expr.body().accept(this, outer);
     }
     var prelude = DynamicSeq.of(Doc.styled(KEYWORD, Doc.symbol("\\")),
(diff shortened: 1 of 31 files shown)


base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -50,7 +50,7 @@ public ConcreteDistiller(@NotNull DistillerOptions options) {
   }
 
   @Override public Doc visitLam(Expr.@NotNull LamExpr expr, Outer outer) {
-    if (!options.showImplicitPats() && !expr.param().explicit()) {
+    if (!(boolean) options.map.get(DistillerOptions.Key.ShowImplicitPats) && !expr.param().explicit()) {
```
ShreckYe, line 53: is this cast necessary? (similar cases below and in `CoreDistiller`)

ShreckYe (commented):
Thanks for requesting. It looks basically fine to me. However, I don't think I am familiar enough with this part of the code to approve it.

However, I don't understand why it's necessary to refactor the options with enums and make `DistillerOptions` optional.

ice1000:
> And I don't understand why it's necessary to refactor the options with enums and make DistillerOptions mutable.

Basically you need to change them, so mutable

Enum means you can have code completion in the REPL for free (no effort) thanks to @imkiva's completion framework

base/src/main/java/org/aya/distill/ConcreteDistiller.java
ice1000: ??? wait what

ice1000:
> is this cast necessary? (similar cases below and in `CoreDistiller`)

Thanks for pointing this out! Fixed in new commit.

imkiva (approved):
bors r+

PR #169, opened 2021-11-11
Print `new` expressions as blocks

### LINKED ISSUE #28: `new` expressions are pretty printed in a very ugly way
So we should improve it

ice1000:
Fix #28


ice1000:
bors merge

ice1000:
Much better
image

ice1000:
Hahah

image

Jacoco slower than bors

PR #176, opened 2021-11-13
Pattern matching refactoring

ice1000:
Fix #170: now we resolve patterns actually in the resolver.

PR #177, opened 2021-11-13
Delete `ExprRefSubst`

### LINKED ISSUE #175: Move some classes from `api` to `tools`
I propose the following:

+ `api` should be API of the **Aya compiler**
+ General compiler utilities go to `tools`
+ `api` depends on `tools`

So we need to move things like `SourcePos`, `Reporter`, `Problem` to `tools`.

ice1000:
Follow-up #176 
Fix #175: move `SourcePos`, `WithPos`, `SourceFile`, etc.
bors merge


ice1000:
bors merge

PR #181, opened 2021-11-14
Infixable patterns

imkiva:


base/src/main/java/org/aya/concrete/parse/AyaProducer.java
```diff
@@ -615,25 +619,17 @@ private static SourcePos sourcePosForSubExpr(@NotNull SourceFile sourceFile, Seq
     return visitAtomPatterns(ctx.atomPatterns()).apply(true, null);
   }
 
-
   public BiFunction<Boolean, LocalVar, Pattern> visitAtomPatterns(@NotNull AyaParser.AtomPatternsContext ctx) {
     var atoms = ctx.atomPattern().stream()
       .map(this::visitAtomPattern)
       .collect(ImmutableSeq.factory());
-    if (atoms.sizeEquals(1)) return (ex, as) -> atoms.first().apply(ex);
+    if (atoms.sizeEquals(1)) return (ex, as) -> newBinOPScope(atoms.first().apply(ex), as);
```
ice1000, line 626: What's the point of one-arg binopseq?

ice1000 (commented):
Is `ErrorPattern` necessary?

base/src/main/java/org/aya/concrete/parse/AyaProducer.java
ice1000: I see now. It's for section parsing.

ice1000: Although we do not want to support sections in patterns, we may. Do it in some day.

ice1000:
bors merge

ice1000 (approved):
lgtm

base/src/main/java/org/aya/concrete/parse/AyaProducer.java
imkiva: It's to keep consistent with expr binopseq since they share the same name and the same pattern of parsing.
We may treat pattern section as underline pattern (like `x:` means `x:_`) (just random thoughts)

imkiva:
> Is `ErrorPattern` necessary?

Maybe yes. It is invoked in `BinOpParser` when fixity or precedence errors occur for not interrupting the compiler. We may handle error patterns in tycker to apply this idea.

ice1000:
@imkiva No, we don't. We mark a clause with errors with a member variable (you can easily find it in `PatTycker`, I can't recall the name), and what we do is avoid checking its body. I guess we just need to find a way to mark the variable in the resolver. The error will be reported so we shouldn't work on it again in the tycker

ice1000:
@imkiva But indeed that some error patterns do not have corresponding errors

PR #187, opened 2021-11-15
TYCK: add TyckOptions(enable or disable sort system)

tonfeiz:
Some problems:
1. Maybe I'm disabling too much of the sort system
2. Not sure about the meaning of "avoid documentation". So I  just didn't comment and threw TyckExceptions.

I'd be happy to modify it if these are indeed the problems, please give some suggestions, thanks!


ice1000 (changes requested):
Don't throw TyckerException. We consider throwing TyckerException bugs

PR #190, opened 2021-11-16
Only enable overlapping pattern semantics on `overlap` modifier

ice1000:
Addresses the problem raised in #98.

Summary:
+ The title.
+ We mean "reduce according to `overlap` modifier" by: if we have that modifier { try to unfold on each clause anyway }, otherwise { try to unfold, if we see a blocked reduction, fail directly }.
+ When checking conditions, we also reduce according to the `overlap` modifier (`orderIndependent` parameter which is passed everywhere).
+ When checking confluence, if we don't have an `overlap` modifier, we still traverse the classification, but we only generate domination warnings in a different way (if any) (see `firstMatchDomination`).
+ A [red-black tree](https://github.com/aya-prover/aya-dev/blob/281670a088f224e1cfe73678254dc91668f5a356/base/src/test/resources/success/redblack.aya) (no deletion, used to sort a list) showcase.
+ Conditions are always required to be confluent to avoid a bug in Arend:

```arend
\data Adams
  | abel Nat
  | cain Nat Nat \with {
    | zero, a => abel a
    | a, zero => abel (suc a)
  }
\func Adams-id (a : Adams) : Adams
  | abel n => abel n
  | cain n m => cain n m
```

I forgot to open an issue about it, and I'm not in the mood of opening issues.

bors d=imkiva

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -61,6 +61,8 @@ check out [how to][proxy] let gradle use a proxy.
 
 + Questions or concerns are welcomed in the discussion area.
   We will try our best to answer your questions, but please be nice.
++ We welcome nitpicks on error reporting! Please let us know anything not perfect.
+  We have already implemented several user-suggested error messages.
 + Before contributing in any form, please read
   [the contribution guideline](https://github.com/aya-prover/aya-dev/blob/master/.github/CONTRIBUTING.md) thoroughly
   and make sure you understand your responsibilities.
--- a/.github/README.md
+++ b/.github/README.md
@@ -23,7 +23,10 @@ Aya is under active development. Nothing guaranteed! However, we can share some 
   and prove [regularity by computation][regularity] thanks to Arend's type theory.
   We also have the classic cubical-flavored [funExt].
   + We are considering moving to cubical type theory.
-+ Overlapping patterns. Very [useful][oop] in theorem proving.
++ Pattern matching with first-match semantics.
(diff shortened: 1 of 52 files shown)


ice1000:
This PR is now ready for review. I hereby invite @imkiva for her criticisms.

imkiva (approved):
More than excited to see we can finally write RBTrees in aya!!!

Question:
Can we remove domination warnings for order-independent and `overlap`-ping patterns? There's no good reason to say some clauses are really `unreachable`? (the clause from which the result is actually calculated is in the superposition state LOL although the implementation code must be executed sequentially which we don't care)

ice1000:
> Question: Can we remove domination warnings for order-independent and `overlap`-ping patterns? There's no good reason to say some clauses are really `unreachable`? (the clause from which the result is actually calculated is in the superposition state LOL although the implementation code must be executed sequentially which we don't care)

Sometimes a clause _is_ totally covered that you can safely removed. The confluence checker guarantees that no matter which clause we choose, we get the same results, so it's in general better to remove redundant things. Imagine you have redundant code in Java, surely you want intellij to suggest a removal.

Anyways, thanks for the approval.

bors r+

ice1000:
> Build succeeded:
> 
> * [gradle-check](https://github.com/aya-prover/aya-dev/runs/4221167285?check_suite_focus=true)

Thank you!

imkiva:
I see. Thanks for kind explanation ❤️.

PR #191, opened 2021-11-16
Use ImmutableArray in core (#189)

### LINKED ISSUE #189: Use arrays in core instead of `ImmSeq`?
Pros:

+ GraalVM easy interop (see @tsao-chi's comment in https://github.com/aya-prover/aya-prover-proto/pull/654#issuecomment-904359358)
+ `Term[]` is shorter than `ImmutableSeq<Term>`
+ [JEP 405](https://openjdk.java.net/jeps/405) makes pattern matching on an entire array possible, and will benefit the 'match the shape of a definition and make it primitive' feature planned in the initial proposal of Aya
+ Kala has a lightweight wrapper for arrays, so we don't lose many conveniences

Cons:

+ Arrays are slightly less safe than `ImmSeq` due to the subtle mutability
+ `@NotNull Term @NotNull []` is a confusing syntax

/cc @imkiva

imkiva:
fix #189 
Since ImmutbaleArrays are ImmutableSeqs, method parameters are not changed

### DIFF
--- a/base/src/main/java/org/aya/concrete/stmt/Sample.java
+++ b/base/src/main/java/org/aya/concrete/stmt/Sample.java
@@ -3,12 +3,12 @@
 package org.aya.concrete.stmt;
 
 import org.aya.api.error.BufferReporter;
-import org.aya.util.error.SourcePos;
 import org.aya.core.def.Def;
 import org.aya.core.def.UserDef;
 import org.aya.tyck.ExprTycker;
 import org.aya.tyck.StmtTycker;
 import org.aya.tyck.error.CounterexampleError;
+import org.aya.util.error.SourcePos;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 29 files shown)


ice1000:
This is not what I wanted! I'll cp the second commit and create a new pr and ask you for review later. Stay tuned!

ice1000 (commented):
I have demonstrated my proposal in the `arrays` branch. Idk if you like it, but I'm starting to believe that it is a bad idea.

ice1000:
MAYBE NOT!!

This may actually be a good idea. Let's keep this branch updated until we realize its usefulness or uselessness.

ice1000:
I think this should be closed. It is unlikely to be merged, the changes are not easily mergable (expect conflicts -- I think the hoas branch will make it worse), and the functional change commit in this PR is cherry-picked into main, so I believe it is worth closing.

ice1000:
We have https://github.com/aya-prover/aya-dev/pull/191/commits/e076e7969fd5a3c5039f1bd9732b0d9533a01231.patch anyway

PR #193, opened 2021-11-16
Enhance something

ice1000:
Summary:

+ 🍒-picked @imkiva's commit from #191 for projection tycking
+ Added hint for index unification failure
+ Simplified grammar
+ Added a TODO for index unification failure in pat classifier. Basically, if we are splitting with clauses present and we cannot decide if we want to split on one of the constructors, we report an error.

bors r+


ice1000:
> Build succeeded:
> 
> * [gradle-check](https://github.com/aya-prover/aya-dev/runs/4230791141?check_suite_focus=true)

Thanks!

PR #194, opened 2021-11-17
Enhance error report for rbTree

ice1000:
Summary:

+ Added `ArrayUtil` 🍒-picked from `arrays` branch (#192)
+ Now index unification failure in constructor selection (in coverage checker) is taken carefully: if the failure is due to a stuck case and there aren't any catch-all patterns, we report an error. (in case there are no patterns, we do not report error, as this will eventually become either an impossible case or a missing case)
+ Added indexed redblack tree (I blame @dramforever for pushing me to do this, it took me the entire afternoon!! But I also appreciate it as this helped me discovering all these bugs)
+ Since `ownerTele` in `CtorDef` isn't equal to the corresponding `DataDef.telescope()` (so in case core of a concall is missing, we can't retrieve `ownerTele`!! This is impossible yet, but may once become possible). To address this problem, we now store `ownerTele` into `CtorDecl.patternTele`.

https://github.com/aya-prover/aya-dev/blob/54c167ba509d63114df816440fb7e70796b13689/base/src/main/java/org/aya/tyck/StmtTycker.java#L194

This line of code constructs `CtorDef.ownerTele`, which is certainly not always the dataParams.

### DIFF
--- a/tools/src/main/java/org/aya/util/ArrayUtil.java
+++ b/tools/src/main/java/org/aya/util/ArrayUtil.java
@@ -1,0 +1,56 @@
+// Copyright (c) 2020-2021 Yinsen (Tesla) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.util;
+
+import kala.tuple.Tuple;
+import kala.tuple.Tuple2;
+import kala.tuple.Tuple3;
+import org.jetbrains.annotations.NotNull;
+
+import java.lang.reflect.Array;
+import java.util.function.Function;
+
+@SuppressWarnings("unchecked")
+public interface ArrayUtil {
+  static <A> A @NotNull [] map(A @NotNull [] array, Function<A, A> f) {
+    var a = (A[]) Array.newInstance(array.getClass().componentType(), array.length);
+    for (int i = 0; i < array.length; i++) a[i] = f.apply(array[i]);
+    return a;
+  }
+
+  static <A, B> B @NotNull [] map(A @NotNull [] array, B @NotNull [] outTyper, Function<A, B> f) {
+    var b = (B[]) Array.newInstance(outTyper.getClass().componentType(), array.length);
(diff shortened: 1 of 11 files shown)


ice1000:
Since this PR fixes build (which is broken for main), I'll push it to main.

bors merge

tools/src/main/java/org/aya/util/ArrayUtil.java
```diff
@@ -0,0 +1,56 @@
...
+import kala.tuple.Tuple;
+import kala.tuple.Tuple2;
+import kala.tuple.Tuple3;
+import org.jetbrains.annotations.NotNull;
+
+import java.lang.reflect.Array;
+import java.util.function.Function;
+
+@SuppressWarnings("unchecked")
+public interface ArrayUtil {
```
imkiva, line 14: I think we can use `ArraySeq.warp().view()` to implement these methods

ice1000: For traversal ones (like `identical`), yes, but for mapping ones, `ArraySeq.wrap().view()` creates a `SeqView`, but these methods create an array.

I propose deleting them if we decide to close #189 as wontfix.

imkiva: We can discuss #189 later when JEP 405 becomes `Delivered`. And I really don't like manipulating with raw arrays (they give a sense of unsafe) 😭

ice1000: Me either. That's why I changed my mind. Arrays are _very bad_ programming interfaces

ice1000: I'm not using these. I just thought that these _might_ be useful somewhere, so I put then into tools in case we need them in the future... And these are general enough to have a potential usage

ice1000:
For the record: https://docs.google.com/presentation/d/1st0TeiIIe_voZ1mx52N4BqnPuxYALaEMHMjCg6Y40vY/edit#slide=id.gfbba5c4add_1_123

PR #198, opened 2021-11-18
Implement "meta patteriables"

### LINKED ISSUE #81: "Meta patterns"
Patterns might be "solved" somehow.

ice1000:
Fix #81, oh my glob.

## Summary

Split into five sections.

### Refactorings
+ Inlined some functions, like `CoreDistiller::visitMaybeCtorPatterns`. There are some others, I can't remember.
+ For `BaseDistiller::visitCalls`, added a new boolean parameter for implicitness option. This is because sometimes it's `ShowImplicitArgs`, sometimes it's `ShowImplicitPats`.
+ I replaced  all visitors of `Pat` with pattern matching (JEP 406) and deleted `Pat.Visitor` the abstract visitor along with all the `doAccept` methods.

### Behavioural improvements
Now we also `zonk` the types of the patterns after `elabClauses`. We didn't do this before, but it so far didn't cause any problem. I conjecture that patterns' types are never used after their type checking, but I'm not exactly sure (I'm 95% sure). If they really aren't used, we probably don't have to zonk them, and we probably want to remove them for spatial efficiency. /cc @imkiva 

### New features
+ Added new `Pat.Meta` for "unknown patterns", which are translated into `RefTerm.MetaPat` (also new) in `PatToTerm`. The idea behind these two structures are similar to "meta variables" but for patterns.
  + The `zonk` for meta patteriables is `Pat::inline` -- they inline the solutions of meta patteriables. When no solution, we turn them into bind patterns. So, `Pat::inline` is total, thus no need to have an error reporter.
  + In `PatMatcher`, we may encounter problems like "matching a pattern `p` with an instance of `RefTerm.MetaPat`". In this case, we "solve" the meta patteriable as a renamed version of `p`. Yes, the renamer for patterns is also added.
  + After all these, we `inline` the patterns, which is like freezing the meta patteriables. Solved meta patteriables are inlined as their solutions.
+ `CalmFace` patterns are now tycked as meta patterns. They will cause `UnsupportedOperationException` before.
+ Implicitly-generated patterns are now tycked as meta patterns if their types are data types (to avoid `MetaPat` appearing in the types of unifications). They were generated as bind patterns before.

### Assumptions
+ We will only see `RefTerm.MetaPat` in `PatMatcher` in `PatTycker`. Other invocations to `PatMatcher` will not. Therefore, we made the new `LocalCtx` parameter of `PatMatcher` nullable, because we need to modify the `localCtx` in the type checking of patterns.
+ Meta patteriables will only be solved once. Unsure what to do when they are solved more than once. Maybe unify the patterns to refine them further? 🤔

### Tests
Modified the redblack tree showcase with all manual implicit refinement removed. It works perfectly. Now the exact Agda code can be translated to Aya, with every omittable pattern omitted. One exception is that I discovered a case when you match on two `{black}` patterns, you don't need to split a `rbBlack a x b` pattern. I decided to keep this code to make the code shorter afterall.

Sorry for my English.

### DIFF
--- a/base/src/main/java/org/aya/core/pat/PatUnify.java
+++ b/base/src/main/java/org/aya/core/pat/PatUnify.java
@@ -20,13 +20,13 @@ import org.jetbrains.annotations.Nullable;
 public record PatUnify(@NotNull TermSubst lhsSubst, @NotNull TermSubst rhsSubst) {
   private void unify(@NotNull Pat lhs, @NotNull Pat rhs) {
     switch (lhs) {
+      default -> throw new IllegalStateException();
       case Pat.Bind bind -> {
       }
       case Pat.Tuple tuple -> {
         if (rhs instanceof Pat.Tuple tuple1) visitList(tuple.pats(), tuple1.pats());
         else reportError(lhs, rhs);
       }
-      case Pat.Absurd absurd -> throw new IllegalStateException();
       case Pat.Prim prim -> {
         if (!(rhs instanceof Pat.Prim prim1)) reportError(lhs, rhs);
         else assert prim.ref() == prim1.ref();
--- a/base/src/main/java/org/aya/core/pat/Pat.java
+++ b/base/src/main/java/org/aya/core/pat/Pat.java
@@ -46,7 +46,8 @@ public sealed interface Pat extends CorePat {
     return new CoreDistiller(options).visitPat(this, BaseDistiller.Outer.Free);
   }
   @NotNull Pat rename(@NotNull Substituter.TermSubst subst, @NotNull LocalCtx localCtx, boolean explicit);
-  @NotNull Pat zonk(@Nullable Zonker zonker);
+  @NotNull Pat zonk(@NotNull Zonker zonker);
+  @NotNull Pat inline();
   void storeBindings(@NotNull LocalCtx localCtx);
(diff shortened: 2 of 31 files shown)


base/src/main/java/org/aya/core/pat/PatUnify.java
```diff
@@ -20,13 +20,13 @@
 public record PatUnify(@NotNull TermSubst lhsSubst, @NotNull TermSubst rhsSubst) {
   private void unify(@NotNull Pat lhs, @NotNull Pat rhs) {
     switch (lhs) {
+      default -> throw new IllegalStateException();
```
imkiva, line 23: Will this `default` dominate other `case`s?

imkiva, line 23: Will the `default` dominate other `case`s?

ice1000: If it does the confluence checker will explode instantly after startup so I guess the answer is 'no' :D

ice1000: bors d=imkiva

imkiva:
bors merge

ice1000:
> bors merge

Remember to resolve all conversations before merging as they are hard requirements

PR #210, opened 2021-11-20
Remove `TyckerException`

tonfeiz:
Some problems:
1. There is still one `TyckerException` left, because I don't know how to enter that path...
2. `TypeMismatchError` is almost the same as `BadTypeError` except for the type of the `desired` param, but I can't figure out a better way to do it...


ice1000 (commented):
Did you format your code

tonfeiz:
> Did you format your code

I suppose yes. Which ones do you think are not formated?

base/src/main/java/org/aya/tyck/error/FieldProblem.java
```diff
@@ -53,4 +54,20 @@ record UnknownField(
...
+    @NotNull FieldDef fieldDef,
+    int supplied
+    ) implements FieldProblem {
+    @Override public @NotNull Doc describe(@NotNull DistillerOptions options) {
+      return Doc.sep(Doc.english("Expected"),
+        Doc.plain(String.valueOf(fieldDef.ref().core.selfTele.size())),
+        Doc.english("args, but found"),
+        Doc.plain(String.valueOf(supplied)),
+        Doc.english("args for field"),
+        Doc.plain(fieldDef.ref().name())
```
ice1000, line 69: Use `linkRef` (see distillers)

base/src/main/java/org/aya/tyck/pat/PatternProblem.java
```diff
@@ -98,6 +98,26 @@ record TooManyPattern(@Override @NotNull Pattern pattern, @NotNull Term retTy) i
     }
   }
 
+  record NotEnoughPattern(@Override @NotNull Pattern pattern, @NotNull Term.Param param) implements PatternProblem {
+    @Override public @NotNull Doc describe(@NotNull DistillerOptions options) {
+      return Doc.vcat(
+        Doc.english("There is no pattern for the param"),
```
ice1000, line 104: parameter

base/src/main/java/org/aya/tyck/pat/PatternProblem.java
```diff
@@ -98,6 +98,26 @@ record TooManyPattern(@Override @NotNull Pattern pattern, @NotNull Term retTy) i
...
+        Doc.par(1, param.toDoc(options)),
+        Doc.english("to match against"));
+    }
+  }
+
+  record TooManyImplicitPattern(@Override @NotNull Pattern pattern, @NotNull Term.Param param) implements PatternProblem {
+    @Override public @NotNull Doc describe(@NotNull DistillerOptions options) {
+      return Doc.vcat(
+        Doc.english("There are too many implicit patterns: "),
```
ice1000, line 113: No need the last whitespace

base/src/test/resources/failure/patterns/not-enough-pat.aya.txt
```diff
@@ -0,0 +1,13 @@
+In file $FILE:4:9 ->
+
+  2 | 
+  3 | def ifElse {A : Type} (b : Bool) (x y : A) : A
+  4 |  | true, x => x
+               ^^
+
+Error: There is no pattern for the param
+         (y : A)
+       to match against
```
ice1000, line 10: We usually say matching a pattern against a parameter

ice1000 (commented):
Incomplete

ice1000:
> > Did you format your code
> 
> I suppose yes. Which ones do you think are not formated? 

I was asking. Probably it's my mobile client not typesetting your code well lol. Small screen sucks

base/src/main/java/org/aya/tyck/error/FieldProblem.java
tonfeiz: `linkDef`, right?

tonfeiz: Seems `linkRef` is not public.

tonfeiz:
> > > Did you format your code
> > 
> > 
> > I suppose yes. Which ones do you think are not formated?
> 
> I was asking. Probably it's my mobile client not typesetting your code well lol. Small screen sucks

So hard-working. I would never use a mobile phone to review code lol.

base/src/main/java/org/aya/tyck/error/TypeMismatchError.java
```diff
@@ -0,0 +1,42 @@
...
+import org.aya.api.error.ExprProblem;
+import org.aya.api.util.NormalizeMode;
+import org.aya.concrete.Expr;
+import org.aya.core.term.Term;
+import org.aya.pretty.doc.Doc;
+import org.jetbrains.annotations.NotNull;
+
+// Difference from BadTypeError: desired type is a term rather than Doc
+// TODO: Maybe there is better ways to deal with this?
+public record TypeMismatchError(
```
imkiva, line 13: What about making `BadTypeError.desired` an `AyaDocile`?

ice1000:
> `linkDef`, right? 

linkRef

base/src/main/java/org/aya/tyck/error/TypeMismatchError.java
tonfeiz: Great! But what about `Docile` instead of `AyaDocile`? `Doc` doesn't extend `AyaDocile`, which makes much original code incompatible. Besides, there isn't much difference to me anyway.

tonfeiz:
> > `linkDef`, right?
> 
> linkRef

Okay, then I'll make it public.

base/src/main/java/org/aya/tyck/error/TypeMismatchError.java
tonfeiz: No, I'm wrong. The problem seems just lies in the difference between `AyaDocile` and `Docile`.  `Doc.xxx` can't be used anymore if using `AyaDocile`, which breaks the original code :sob: .

imkiva: > `Doc.xxx` can't be used anymore if using `AyaDocile`

Maybe we should generate `Doc` from `AyaDocile` with `AyaDocile.toDoc` first

imkiva: And there's a simple way to make `Doc` act like an `AyaDocile`: replace `Doc.xxx()` with `options -> Doc.xxx()`

tonfeiz: Wow, that's amazing. I'll try that. Thanks a lot!

imkiva (approved):
Nice work! 

For the last `TyckerException` in
https://github.com/aya-prover/aya-dev/blob/491539bd232f4c9760c518291be708b1e477d39b/base/src/main/java/org/aya/tyck/ExprTycker.java#L346-L350

I think it's unreachable for `SigmaExpr` since it is guaranteed by `AyaProducer` that the `type` field is never `null` (in this case, a `HoleExpr` is created), see

https://github.com/aya-prover/aya-dev/blob/f1775a6abbe15b5201e8956e9a3564b544dd6384/base/src/main/java/org/aya/concrete/parse/AyaProducer.java#L601-L605

(Just find usage of `SigmaExpr` to see how it is `new`-ed. Maybe incorrect)

ice1000:
> > Did you format your code
> 
> I suppose yes. Which ones do you think are not formated?

image

base/src/main/java/org/aya/tyck/error/BadTypeError.java
```diff
@@ -24,42 +25,50 @@ public record BadTypeError(
...
       Doc.english("you gave"),
-      Doc.english("struct type"));
+      options -> Doc.english("struct type"));
+  }
+
+  public static @NotNull BadTypeError lamParam(@NotNull Expr lamExpr, @NotNull AyaDocile paramType,
+                                               @NotNull Term actualParamType) {
+    return new BadTypeError(lamExpr, actualParamType,
+      Doc.english("apply or construct"),
+      Doc.english("of the lamda's param"),
```
ice1000, line 71: ..

ice1000:
bors r=[ice1000, imkiva]

ice1000:
Do it again

bors r=imkiva,ice1000

PR #217, opened 2021-11-22
Introduce `TyckUnit`

ice1000:
Prepare for some future work

### DIFF
--- a/base/src/main/java/org/aya/concrete/stmt/Signatured.java
+++ b/base/src/main/java/org/aya/concrete/stmt/Signatured.java
@@ -16,7 +16,7 @@ import org.jetbrains.annotations.Nullable;
  *
  * @author ice1000
  */
-public sealed abstract class Signatured implements ConcreteDecl, OpDecl permits Decl, Decl.DataCtor, Decl.StructField {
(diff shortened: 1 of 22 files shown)


base/src/main/java/org/aya/concrete/stmt/Signatured.java
```diff
@@ -16,7 +16,7 @@
  *
  * @author ice1000
  */
-public sealed abstract class Signatured implements ConcreteDecl, OpDecl permits Decl, Decl.DataCtor, Decl.StructField {
+public sealed abstract class Signatured implements ConcreteDecl, OpDecl, org.aya.tyck.order.TyckUnit permits Decl, Decl.DataCtor, Decl.StructField {
```
imkiva, line 19: Why use qualified class names?

base/src/main/java/org/aya/concrete/stmt/Stmt.java
```diff
@@ -16,17 +16,15 @@
...
-import org.aya.util.error.SourcePos;
+import org.aya.util.error.SourceNode;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * @author kiva
  */
-public sealed interface Stmt extends AyaDocile
+public sealed interface Stmt extends AyaDocile, SourceNode, org.aya.tyck.order.TyckUnit
```
imkiva, line 26: ditto

PR #219, opened 2021-11-22
Try to fix some obvious bugs of the library system

ice1000:
Follow-up #212

PR #221, opened 2021-11-22
An alternative solution to #219

ice1000:
See #219. Mutually exclusive


ice1000:
[cannot-parse.zip](https://github.com/aya-prover/aya-dev/files/7583805/cannot-parse.zip)

ice1000:
Cherry-picked to #219

Issue #144, opened 2021-11-05
Merge or delete tests for already stabilized features (like parser, etc.)


imkiva:
like `emoji-*-.aya`, `syntax.aya`

ice1000:
Duplicate of #29

imkiva:
I mean not only success tests. There are lots of failure tests that make the development harder, for example:
- syntax errors. These tests' `.aya.txt` need to be regenerated every time we change the grammar.
- files with multiple definitions for testing on the same thing. The order of problems reported is not guaranteed

ice1000:
> I mean not only success tests. There are lots of failure tests that make the development harder, for example:
> - syntax errors. These tests' `.aya.txt` need to be regenerated every time we change the grammar.
> - files with multiple definitions for testing on the same thing. The order of problems reported is not guaranteed (maybe it is not a problem of tests)

I agree with you, and I can see the problems too. But how to address the problem?

PR #224, opened 2021-11-23
Fix more bugs related to library system

imkiva:
What we have missed before has been implemented now including:

- bind graph is not imported when importing another module
- reporter is not cleared so the compiler stops writing successful core files
- forgot to set counterexample statements' module
- #29
- #144 

With new refactoring:
- generalize `NonStoppingTycker` (named `IncrementalTycker` before)


ice1000:
image

ice1000:
Insanely productive!!!
bors r+

PR #229, opened 2021-11-23
Improve cli output, add `--verbosity` flag

ice1000:
As title

### DIFF
--- a/base/src/test/java/org/aya/test/LibraryTest.java
+++ b/base/src/test/java/org/aya/test/LibraryTest.java
@@ -4,18 +4,31 @@ package org.aya.test;
 
 import org.aya.cli.library.LibraryCompiler;
 import org.aya.util.FileUtil;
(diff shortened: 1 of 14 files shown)


base/src/test/java/org/aya/test/LibraryTest.java
```diff
@@ -4,18 +4,31 @@
...
 import java.io.IOException;
+import java.nio.file.Path;
 
 public class LibraryTest {
   @Test public void test() throws IOException {
-    main();
+    compile();
+    // Run twice, the second time should load the cache directly.
+    compile();
   }
```
imkiva, line 19: Try run 3 times:
- build all from source
- Remove test out, build test from source, while reusing the common cache
- build all from cache

ice1000: 🤔

PR #230, opened 2021-11-23
Run library test for 3 times as said in the review

ice1000:
https://github.com/aya-prover/aya-dev/pull/229#discussion_r754967702


imkiva (approved):
bors r+

Issue #228, opened 2021-11-23
Reuse Antlr parse tree instead of 'only parse imports'


ice1000:
Because even if we only get imports, we still reparse everything. It's better to reuse the parse tree

PR #231, opened 2021-11-24
Reuse parsed tree instead of import only

imkiva:
A simple fix of #228 

This PR adds two field to `LibrarySource`
- `Ref<ImmutableSeq<Stmt>>` as concrete trees which is initialized after resolving library imports
- `Ref<ResolveInfo>` which I don't know whether can be reused by LSP (if not, we remove it then)


imkiva:
Coverage decreased? but how can it be?

ice1000 (commented):
What's the relationship between cached library source and library source?

PR #235, opened 2021-11-24
Rely on `LocalCtx`

ice1000:
As title, following #234


ice1000:
Now we depend less on `RefTerm::type`

PR #237, opened 2021-11-25
A step towards mixfix parser

imkiva:
This is a very simple mixfix parser that is based on the generalized BinOP parser. 
We only expect the mixfix operator to be used in the situation where we cannot place an infix operator, like

- bracket overloading, like `[_]` or `[_,_]`
- common syntax, like `if_then_else_` or `_[_]`

Some features are still missing but will be done in the future.
See [Binop.aya](https://github.com/aya-prover/aya-dev/blob/mixfix/base/src/test/resources/success/src/Binop.aya) for what is supported and what is not


ice1000 (approved):
This is beautiful!!

ice1000:
Syntactically, `if_then_else_` is different from `if_then_else`, but your implementation treats them the same way. Have you considered my proposal upon a different syntax for name parts? Like `def mixfix {if then else}`, which eliminates such difference.

imkiva:
> Syntactically, `if_then_else_` is different from `if_then_else`, but your implementation treats them the same way. Have you considered my proposal upon a different syntax for name parts? Like `def mixfix {if then else}`, which eliminates such difference.

Yes, the underscore should be removed in the next commit. I am testing abused mixfix cases and trying to disallow them

ice1000:
@imkiva Cool!!! 🤩🤩

imkiva:
I don't think removing underscore is a good idea. For example, `_[_]` becomes `{ [ ] }` when underscores are removed, which is confusing and harder to read and recognize its usage.
Maybe we need another solution 🤔

ice1000:
> I don't think removing underscore is a good idea. For example, `_[_]` becomes `{ [ ] }` when underscores are removed, which is confusing and harder to read and recognize its usage. Maybe we need another solution 🤔

I agree with that. I think underscores are kinda fine, as I rarely see people getting mad at Agda syntax anyways

ice1000:
bors d=imkiva

imkiva:
I think the mixfix parser in this PR may add more confusing than agda's. I will come up with a better solution some day. Closing now.

imkiva:
But some useful refactors can be cheery-picked.

imkiva:
https://github.com/aya-prover/aya-dev/pull/237.patch

ice1000:
> https://github.com/aya-prover/aya-dev/pull/237.patch

What do you mean?

imkiva:
> > https://github.com/aya-prover/aya-dev/pull/237.patch
> 
> What do you mean?

In case someday I will need it because I deleted this branch locally and cloudly just a few hours ago.

ice1000:
> > > https://github.com/aya-prover/aya-dev/pull/237.patch
> > 
> > 
> > What do you mean?
> 
> In case someday I will need it because I deleted this branch locally and cloudly just a few hours ago.

You decide to abandon this feature? I still find it useful somehow

imkiva:
> You decide to abandon this feature?

No. I only decide to abandon this implementation

PR #244, opened 2021-11-30
Get rid of RefTerm.type

imkiva:
This should be done after #235 I suppose

Not sure if the `TermToPat` is still needed because I cannot find any usage of it.


ice1000:
Term to Pat will be used in the future

ice1000:
But I'm quite surprised that we can actually delete it for now. We may recover the term to lhs in the future so this is actually fine!

bors r+ let's get this in

PR #245, opened 2021-12-01
Library in LSP

imkiva:
Summary

- Decoupled LibraryCompiler and LibraryOwner (also needed by #226).
- Refactored LSP to only work with the library system.
- Auto-detect library root and load at startup.
- Generate mocked library for the single source file (like rust-playground emacs plugin).
- Go to definition across modules.

Depends on aya-prover/aya-vscode#18


base/src/main/java/org/aya/concrete/parse/AyaParsing.java
```diff
@@ -76,7 +76,7 @@
     @NotNull Reporter reporter, @NotNull String text,
     @NotNull BiFunction<AyaProducer, AyaParser, T> tree
   ) {
-    var sourceFile = new SourceFile(Path.of("stdin"), text);
+    var sourceFile = new SourceFile("<stdin>", Path.of("stdin"), text);
```
ice1000, line 79: Please, don't do this. On Windows, it'll say that it's an invalid path.

base/src/main/java/org/aya/concrete/visitor/StmtConsumer.java
```diff
@@ -31,7 +31,7 @@ default void visitClause(@NotNull Pattern.Clause c, P pp) {
 
   default void visitPattern(@NotNull Pattern pattern, P p) {
     switch (pattern) {
-      case Pattern.Ctor ctor -> ctor.params().forEach(param -> visitPattern(pattern, p));
+      case Pattern.Ctor ctor -> ctor.params().forEach(pat -> visitPattern(pat, p));
```
ice1000, line 34: Thanks!!!! Lmfao

lsp/src/main/java/org/aya/lsp/library/WsLibrary.java
```diff
@@ -0,0 +1,73 @@
...
+ * @author kiva
+ */
+public record WsLibrary(
+  @NotNull CountingReporter reporter,
+  @NotNull SourceFileLocator locator,
+  @NotNull DynamicSeq<LibrarySource> sources,
+  @NotNull LibraryConfig mockConfig,
+  @NotNull Path workspace
+) implements LibraryOwner {
+  public static @NotNull WsLibrary from(@NotNull Reporter outReporter, @NotNull Path ayaSource) {
```
ice1000, line 31: I propose renaming to `mock`

tools/src/main/java/org/aya/util/error/SourceFile.java
```diff
@@ -6,23 +6,20 @@
...
+  @NotNull String sourceCode
+) {
+  public SourceFile(@NotNull String display, @NotNull Path underlying, @NotNull String sourceCode) {
+    this(display, Option.some(underlying), sourceCode);
   }
 
-  public static final SourceFile NONE = new SourceFile(Option.none(), "");
-  public static final SourceFile SER = new SourceFile(Option.none(), "");
+  public static final SourceFile NONE = new SourceFile("<unknown-file>", Option.none(), "");
+  public static final SourceFile SER = new SourceFile("<serialized-core>", Option.none(), "");
```
ice1000, line 20: Ditto about path names

ice1000 (commented):
Does WsLibrary work with single file opened as a project?

ice1000:
> Does WsLibrary work with single file opened as a project?

I think the answer is already in the summary. I was dumb

base/src/main/java/org/aya/concrete/parse/AyaParsing.java
imkiva: Not anymore. The display field is just a plain `String` now :)

base/src/main/java/org/aya/concrete/visitor/StmtConsumer.java
imkiva: Copilot did it. 🤣

tools/src/main/java/org/aya/util/error/SourceFile.java
imkiva: They will not be used to construct a `Path`, so it is safe to keep brackets.

imkiva: Instead, we store the real path in `SourceFile#underlying`

base/src/main/java/org/aya/concrete/parse/AyaParsing.java
ice1000: Noice.

tools/src/main/java/org/aya/util/error/SourceFile.java
ice1000: Noice

imkiva:
> Btw have you resolved the problem about the 'triggering exception in compiler breaks highlighting' problem?

Not yet. Maybe a single try-catch would do

PR #253, opened 2021-12-02
Fix prim levels

ice1000:
Improve `common`. Probably I'm gonna add more code in this PR stay tuned


ice1000:
Reference: https://github.com/JetBrains/arend-lib/blob/3cbb505dbc8bf2cc149be3084f80db7426fed587/test/Data/Vec/Properties.vc

PR #255, opened 2021-12-03
Enhance reporters

### LINKED ISSUE #223: Respect user options in cli-reporter?
https://github.com/aya-prover/aya-dev/blob/b967c39e7ca12a93670c89e165d8ee7f8a9533aa/cli/src/main/java/org/aya/cli/single/CliReporter.java#L28

But this could break the tests.

imkiva:
- close #223
- make `CountingReporter` an interface
-  make some output styled

### DIFF
--- a/pretty/src/main/java/org/aya/pretty/backend/string/custom/UnixTermStyle.java
+++ b/pretty/src/main/java/org/aya/pretty/backend/string/custom/UnixTermStyle.java
@@ -11,4 +11,6 @@ public enum UnixTermStyle implements Style.CustomStyle {
   Blink,
   Overline,
   Reverse,
+  TerminalRed,
+  TerminalGreen,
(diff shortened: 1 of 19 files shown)


pretty/src/main/java/org/aya/pretty/backend/string/custom/UnixTermStyle.java
```diff
@@ -11,4 +11,6 @@
   Blink,
   Overline,
   Reverse,
+  TerminalRed,
+  TerminalGreen,
 }
```
imkiva, line 16: These are not like `Style.color(0xFF0000)` which set the absolute color, instead, they are decided by the terminal (like user-customized color scheme)

ice1000 (approved):
bors r+

PR #265, opened 2021-12-06
Delay equations on flex-flex cases

### LINKED ISSUE #160: Scope checker misled by unused `contextArgs`
My temporary solution is to inline the meta solutions so the `contextArgs` are cleared out, but we may need to _prune_ the context somehow (if we solve a meta as a term that contains metas with smaller scopes, we may restrict the inner metas' scopes to be the same as the outers, and let it go).

ice1000:
Fix #160


ice1000:
bors merge

PR #270, opened 2021-12-07
Special case for simple functions

### LINKED ISSUE #264: Nest lambda expression and implicit type annotation in func definition
Recently I plan to translate some code in `PLFA` and `Lean`'s tutorial into `Aya`. I believe this is helpful to my understanding of `Aya` and it can also be seen as a test for `Aya`. These may not be serious issues and maybe low priority, but I'll post them for now anyway. 

The following code doesn't type check:
`def fun1 => \x => \y => ifElse (not y) (x + suc zero) (x + suc (suc zero))`

Do we support nest lambda expressions now? 

Another one is:
`def fun3 x => ifElse true (x + suc zero) (x + suc (suc zero))`
The error msg is :
```
40 | def fun3 x => ifElse true (x + suc zero) (x + suc (suc zero))
Error: The name `x` is not defined in the current scope 
```

I'm not sure where the problem is in this one. BTW, if this is out of expectation, I can try to fix it.

ice1000:
Fix #264 but doesn't yet pass test due to a 离谱 case

https://www.bilibili.com/video/BV16U4y1t7wH/

### DIFF
--- a/api/src/main/java/org/aya/api/error/Problem.java
+++ b/api/src/main/java/org/aya/api/error/Problem.java
@@ -13,6 +13,7 @@ import org.aya.pretty.doc.Doc;
 import org.aya.pretty.doc.Style;
 import org.aya.pretty.doc.Styles;
 import org.aya.pretty.error.PrettyError;
+import org.aya.pretty.style.AyaStyleFamily;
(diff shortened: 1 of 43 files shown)


imkiva:
bors d=ice1000

PR #277, opened 2021-12-10
Rewrite `DefEq` by separating the `lr` map and the `rl` map

### LINKED ISSUE #275: Refactor `DefEq` to make it work
https://github.com/aya-prover/aya-dev/blob/a30b95f4ed2a7b4c71a1ad028384600aab866a59/base/src/test/resources/success/common/src/Data/Vec.aya#L28

ice1000:
It is obvious. Fixes #275. Don't merge yet.


ice1000:
Ohhh!! I was trying to make a program compile, but it turns out that the program was wrong

Issue #242, opened 2021-11-28
Remove `type` in `Pat`


tonfeiz:
It seems there are still some places where some specific `Pat`'s types are useful, like in
https://github.com/aya-prover/aya-dev/blob/85e41c5f5098a5685d007f6000f305ad9703397e/base/src/main/java/org/aya/core/pat/PatUnify.java#L49-L53
where `Pat.Bind`'s type is used.

`Pat.Meta` and `Pat.Ctor`'s types are also used, and seems can not be deleted in my opinion(may be incorrect).

So I'm wondering do you mean to delete all types in `Pat`, or maybe save some of them? @ice1000

ice1000:
> So I'm wondering do you mean to delete all types in Pat, or maybe save some of them? @ice1000

I meant all. `PatUnify` is only used in the tycking of patterns, not the evaluation of patterns. Therefore can be dropped _after_ the tycking.

My expectation would be to introduce a 3rd AST for patterns (I propose the interface name `LHS`), which is a version of `Pat` without types and absurd/Meta cases

tonfeiz:
OK, I understand it now. Thanks a lot.

ice1000:
Well. First of all, do we need types of ctor/tuple patterns? Maybe we can remove these first. It should be acceptable to have only bind patterns typed

tonfeiz:
That should be OK. Then I'll try to implement a version of this first.

tonfeiz:
I've met two problems in doing this, and I want to make sure what I intend to do meets your expectation. @ice1000 

1. I remember that you want to replace `Pat` in `Matching` with `Lhs`, however `Conquer` also uses `Matching` and there's also unification process, so I think maybe it's also a type checking process. If so I would make a new `TyckedMatching` class and use it in `Def`s. 
2. `PatMatcher` needs to be refactored to `LhsMatcher`, but as I said it will also be used in type checking process, so it should also accept `Pat` as its argument. Now it shoud accept two types of arguments with almost the same logic. I don't think it's a good idea to implement the logic twice, but I also wonder whether transfer `Lhs` to `Pat` is a good idea because logically speaking it shouldn't happen. 

Sorry it's a bit long, but I think asking in advance is better than fixing it afterwards. :rofl:

ice1000:
1. Is Conquer unifying patterns?
2. I don't know. Maybe you can generate the underlying Lhs of a Pat earlier and use Lhs for matching only

ice1000:
[...]

Issue #271, opened 2021-12-08
MetaPat bug?


ice1000:
```aya
def infix !! (l : List A) (i : Fin (length l)) : A
| a cons l, fzero => a
| a cons l, fsuc i => l !! i
```

Works but

```aya
def infix !! (l : List A) (i : Fin (length l)) : A
| a cons _, fzero => a
| a cons l, fsuc i => l !! i
```

is rejected for the following reason

![image](https://user-images.githubusercontent.com/16398479/145166295-22d74b64-df1f-4839-bbe4-ed0b3f356c2e.png)

ice1000:
Also the pattern pp is not doing a gj

ice1000:
```
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'localCtx' of org/aya/core/pat/Pat$Ctor.rename must not be null
	at org.aya/org.aya.core.pat.Pat$Ctor.$$$reportNull$$$0(Pat.java)
	at org.aya/org.aya.core.pat.Pat$Ctor.rename(Pat.java)
	at org.aya/org.aya.core.pat.PatMatcher.solve(PatMatcher.java:102)
	at org.aya/org.aya.core.pat.PatMatcher.match(PatMatcher.java:77)
	at org.aya/org.aya.core.pat.PatMatcher.match(PatMatcher.java:111)
	at org.aya/org.aya.core.pat.PatMatcher.tryBuildSubstTerms(PatMatcher.java:49)
	at org.aya/org.aya.core.pat.PatMatcher.tryBuildSubstArgs(PatMatcher.java:39)
	at org.aya/org.aya.core.visitor.Unfolder.tryUnfoldClauses(Unfolder.java:118)
	at org.aya/org.aya.core.visitor.Unfolder.tryUnfoldClauses(Unfolder.java:108)
	at org.aya/org.aya.core.visitor.Unfolder.visitFnCall(Unfolder.java:77)
	at org.aya/org.aya.core.visitor.Unfolder.visitFnCall(Unfolder.java:33)
	at org.aya/org.aya.core.term.CallTerm$Fn.doAccept(CallTerm.java:58)
	at org.aya/org.aya.core.term.Term.accept(Term.java:44)
	at org.aya/org.aya.core.visitor.TermFixpoint.visitArg(TermFixpoint.java:108)
	at org.aya/org.aya.core.visitor.Unfolder.lambda$visitConCall$1(Unfolder.java:48)
[...]

PR #281, opened 2021-12-11
LHS: add Lhs as type checked Pat

tonfeiz:
See #242.

There are some problems:
1. Many duplicate code with `Pat` in serialize and distiller process, but seems no better ways
2.  `PatCCTest` cannot be passed because we relied on `Pat` in `Def` to classify, but now we're using `Lhs`. I'm still thinking about better ways to deal with that. 
3. `PatMatcher` has been changed a lot(maybe not a lot, but the logic is kind of different than before anyway), so pay attention to this when reviewing.

I may add some comments or tests afterwards, but the overall code won't be changed a lot, so I just push them for now. I'm pretty sure there must be some(or many :rofl:) problems, just comment and I'll fix them.


tonfeiz:
Another problem is `type` for `Lhs.Ctor` is saved because in `LhsToTerm` we need to generate a `CallTerm.Con`. It seems necessary.

ice1000 (changes requested):
Branch out of date of main

ice1000:
> 1. Many duplicate code with `Pat` in serialize and distiller process, but seems no better ways

Remove the serialization code for `Pat` plz.

> 2. `PatCCTest` cannot be passed because we relied on `Pat` in `Def` to classify, but now we're using `Lhs`. I'm still thinking about better ways to deal with that.

Yes, you forgot to replace the `Pat` in `Def` with `Lhs`.

> 3. `PatMatcher` has been changed a lot (maybe not a lot, but the logic is kind of different than before anyway), so pay attention to this when reviewing.

You forgot to deal with `RefTerm.MetaPat`.

> I may add some comments or tests afterwards, but the overall code won't be changed a lot, so I just push them for now. I'm pretty sure there must be some (or many 🤣) problems, just comment and I'll fix them.

I've pushed some, you may wanna take a look. Remember to rebase on latest main tho.

tonfeiz:
> You forgot to deal with `RefTerm.MetaPat`.

No, I've dealt with that. However it's different than before. Just take a closer look at it.



> Yes, you forgot to replace the `Pat` in `Def` with `Lhs`.

That's not the problem here :rofl:. It's just because I've replaced `Pat`  with `Lhs` in `Def`, so we can't writing code like in `PatCCTest` anymore. `PatCCTest` is kind of like cheating, we first type checked some code, then we use the type checked `Pat` to classify. Now we are using `Lhs`, which prevents this kind of cheating, but also break the code and makes it really difficult to rewrite.

For your question, I know there are still `Pat` in `CtorDef`. However, it seems we can't replace it with `Lhs` because of #271. When we `selectCtor` in `PatTycker`, we'll use `Pat` in `CtorDef`, and perform the matching process, while at the same time we also expect to solve some `MetaPat`. If we change `Pat` to `Lhs`, then `MetaPat`'s solution might be an `Lhs`!

Another negative result of that is we cannot remove serialization code for `Pat` as it's still used in `CtorDef`.

To be honest, I'm kind of wondering is it really necessary to do the refactoring? I mean,  `MetaPat` is kind of blurring the boundary between `Lhs` and `Pat`. And refactoring is the reverse process, which adds some more subtleties.

ice1000:
> No, I've dealt with that. However it's different than before. Just take a closer look at it.

😭

> `PatCCTest` is kind of like cheating, we first type checked some code, then we use the type checked `Pat` to classify. Now we are using `Lhs`, which prevents this kind of cheating, but also break the code and makes it really difficult to rewrite.

It tests the shape of the classification results. I guess you can delete it if it's annoying, as pattern classification is a rather stable feature (future refactoring won't change this part of the algorithm either).

> If we change `Pat` to `Lhs`, then `MetaPat`'s solution might be an `Lhs`!

Oh no, that's beyond my estimation! 😟 What about this: we remove `Pat::type` and `Pat.Tuple::type`, and make a function from `Lhs` to `Pat`?

> Another negative result of that is we cannot remove serialization code for `Pat` as it's still used in `CtorDef`.

Yes!! I really wanna kill that 😭

> To be honest, I'm kind of wondering is it really necessary to do the refactoring? I mean, `MetaPat` is kind of blurring the boundary between `Lhs` and `Pat`. And refactoring is the reverse process, which adds some more subtleties.

This can help us understand the code better. See? I wasn't aware of the fact that meta patteriable unification requires `Pat` instead of `Lhs`. I've learned from this work, tyvm!

ice1000:
> Oh no, that's beyond my estimation! 😟 What about this: we remove `Pat::type` and `Pat.Tuple::type`, and make a function from `Lhs` to `Pat`?
> 
> >

I've done that in a recent commit: 61ab8803bcd3c691efd9d71f975e8db3dc271eb5

ice1000:
Created #283

tonfeiz:
> we remove `Pat::type` and `Pat.Tuple::type`, and make a function from `Lhs` to `Pat`?

That's OK. However, I should remind you that you're blurring the boundary furthurmore. Now they are interchangeable, and once we add something in `Pat`, we'll have to add the same thing in `Lhs`.

> I've done that in a recent commit: [61ab880](https://github.com/aya-prover/aya-dev/commit/61ab8803bcd3c691efd9d71f975e8db3dc271eb5)

You're too productive :rofl:. I can do the rest(add  `Pat`->`Lhs` and delete `SerPat` and more) later if you're not in a hurry.

ice1000:
> That's OK. However, I should remind you that you're blurring the boundary furthurmore. Now they are interchangeable, and once we add something in `Pat`, we'll have to add the same thing in `Lhs`.

Yes, that's also what I see. Once we have JEP 420 merged, we may use a type tag for distinguishing `Pat` and `Lhs`:

+ `Keep : Tag`
+ `Pat<T extends Tag>`
+ `Ctor<T>(Seq<Pat<T>>) : Pat<T>` and similarly `Tuple`
+ `Bind : Pat<Keep>`
+ `Absurd : Pat<Tag>`
+ `MetaPat : Pat<Tag>`

Then, we can have two `Matching`s, one contains `Pat<Keep>` and one contains `Pat<Tag>` -- the latter may contain `Absurd` or `MetaPat` but the former can't.

ice1000:
I propose closing this, as this is a negatively successful experiment (that we now know that something's impossible).

tonfeiz:
> I propose closing this, as this is a negatively successful experiment (that we now know that something's impossible).

Sure. Just close it~

ice1000:
Closed upon agreement

ice1000:
For the record, the commit are still accessible via GitHub even I deleted the branch: 4de80a4b7fa9e2d8beda0d4a7fd543e8dc9b3816

https://github.com/aya-prover/aya-dev/commit/4de80a4b7fa9e2d8beda0d4a7fd543e8dc9b3816.patch

PR #287, opened 2021-12-12
Termination checker based on foetus dependencies

imkiva:
## Summary
This PR implements a termination checker for recursive functions (other recursive definitions requires more research) based on
- _A Predicative Analysis of Structural Recursion_ by Andreas Abel and Thorsten Altenkirch

and implementation details by various people:
- [agda: Termination.hs](https://github.com/agda/agda/blob/master/src/full/Agda/Termination/Termination.hs)
- [arend: BaseCallGraph.java](https://github.com/JetBrains/Arend/blob/master/base/src/main/java/org/arend/typechecking/termination/BaseCallGraph.java)

## Technical details
The termination checker is based on the idea that a term (value) `a` is structurally smaller than another term (value) `b` if `b` can be evaluated to the form of `C(..., a, ...)` where `C` is a constructor of an inductive data type and a recursive function which calls itself directly or indirectly with structurally smaller argument(s).

In this implementation, we care about 3 cases (mainly the foetus dependencies) that can prove the structural order between two terms: _constructor elimination_ by pattern matching, _applications_ and _projections_.


base/src/main/java/org/aya/terck/CallGraph.java
```diff
@@ -2,31 +2,90 @@
...
+    set.removeAll(existing -> matrix.compare(existing) == Relation.LessThan);
     set.add(matrix);
     return true;
   }
+
+  /** the completion of a call graph is finding its transitive closure */
+  private static <T, P> @NotNull CallGraph<T, P> complete(@NotNull CallGraph<T, P> start) {
+    var oldGraph = new Ref<>(start);
+    var newEdge = new Ref<>(1);
+    // TODO: do we really need this fuel since we are just looking for a fixpoint?
```
ice1000, line 52: What does Agda/Arend do?

ice1000 (commented):
Haven't looked 100%

base/src/main/java/org/aya/terck/CallGraph.java
imkiva: They don't have it. I think the fuel can reduce unnecessary loops because we only need to traverse each edge in this graph (according to Warshall’s algorithm)

base/src/main/java/org/aya/terck/Relation.java
```diff
@@ -32,7 +43,17 @@
     return this.lessThanOrEqual(rhs) ? rhs : this;
   }
 
+  /**
+   * A relation `A`is less than another relation `B` iff:
```
ice1000, line 47:
```suggestion
   * A relation `A` is less than another relation `B` iff:
```

ice1000 (approved):
image

ice1000:
Do we have existing tests that are nonterminating?

ice1000:
One thing I wanna do is to change the order of these checks: we should terck before coverage/confluence check. Now I wonder how can we do that 🤔🤔

ice1000:
I guess I should extract the code for coverage/confluence to a function🤔🤔

ice1000:
Btw, I think by doing so we can drastically simplify `StmtTycker` -- making the code cleaner for reading purposes

ice1000:
bors merge this (I guess I want to understand the code later, but the overall architecture of terck seems portable. I love PRs in this style)

PR #290, opened 2021-12-13
Handle file create and delete event in LSP

### LINKED ISSUE #286: LSP should check library source changes
So we can create new sources in vscode without restarting it

imkiva:
fix #286


ice1000 (approved):
This is more beautiful than I expected it to be

bors r+

PR #293, opened 2021-12-15
Improve pretty printer

ice1000:
Something I've done in #287 but it doesn't feel nice enough. I'm refactoring both the impl and the code design

bors d=imkiva


imkiva (approved):
bors r+

PR #302, opened 2021-12-20
Fix a simple problem

imkiva:
bors merge

### DIFF
--- a/lsp/src/main/java/org/aya/lsp/utils/Resolver.java
+++ b/lsp/src/main/java/org/aya/lsp/utils/Resolver.java
@@ -21,6 +21,7 @@ import org.aya.concrete.visitor.StmtConsumer;
 import org.aya.core.def.DataDef;
 import org.aya.core.def.Def;
 import org.aya.core.def.StructDef;
+import org.aya.generic.Constants;
 import org.aya.util.error.SourcePos;
 import org.aya.util.error.WithPos;
 import org.eclipse.lsp4j.Position;
@@ -156,7 +157,9 @@ public interface Resolver {
     }
 
     @Override public void visitSignatured(@NotNull Signatured signatured, XY xy) {
-      signatured.telescope.forEach(tele -> check(xy, tele.ref(), tele.sourcePos()));
+      signatured.telescope
+        .filterNot(tele -> tele.ref().name().startsWith(Constants.ANONYMOUS_PREFIX))
+        .forEach(tele -> check(xy, tele.ref(), tele.sourcePos()));
       super.visitSignatured(signatured, xy);
     }
 


imkiva:
bors cancel

imkiva:
bors merge

lsp/src/main/java/org/aya/lsp/utils/Resolver.java
```diff
@@ -156,7 +157,9 @@
     }
 
     @Override public void visitSignatured(@NotNull Signatured signatured, XY xy) {
-      signatured.telescope.forEach(tele -> check(xy, tele.ref(), tele.sourcePos()));
+      signatured.telescope
+        .filterNot(tele -> tele.ref().name().startsWith(Constants.ANONYMOUS_PREFIX))
```
ice1000, line 161: What if this name is written by users?

imkiva: We should reserve the meaning of names starting with an underscore. It means _does not matter_. If users any term uses those names the resolver should raise an warning or error.

ice1000: Is this just a performance improvement? What if there's a lambda expression in an anonymous parameter?

imkiva: > Is this just a performance improvement?

Yes, finding usages on each LSP action (like hover) is very slow.

> What if there's a lambda expression in an anonymous parameter?

I don't get it. Care to give an example?

ice1000: > We should reserve the meaning of names starting with an underscore. It means _does not matter_. If any term uses those names the resolver should raise a warning or even an error.

I agree

ice1000: > I don't get it. Care to give an example?

Like `def test (Ok (\ x => x))`, the above code implies that this `Ok (\ x => x)` will not be searched in this visitor. Is this intended?

imkiva: > > I don't get it. Care to give an example?
> 
> Like `def test (Ok (\ x => x))`, the above code implies that this `Ok (\ x => x)` will not be searched in this visitor. Is this intended?

No, the type is always searched

imkiva: This PR fixes searching `PsiNamedElement` alike things, that is an anonymous parameter never introduces a usable name.

ice1000: Oh I see, it's searched in `super.visitSignatured`, tyvm!

ice1000: bors r+

PR #309, opened 2021-12-21
Fix a critical bug of SccTycker

imkiva:
- Fixed not clearing reporter countings when one definition failed to tyck (which causes all coming definitions to be treated as failed)
- Search both core and concrete term when highlighting
- Skip `DefVar`s from skipped modules and libraries

### DIFF
--- a/api/src/main/java/org/aya/api/error/CollectingReporter.java
+++ b/api/src/main/java/org/aya/api/error/CollectingReporter.java
@@ -15,12 +15,4 @@ public interface CollectingReporter extends CountingReporter {
   @Override default void clear() {
     problems().clear();
   }
-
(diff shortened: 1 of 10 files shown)


imkiva:
Suggest: replace all `@UnknownNullability` with `@Nullable` to force the IDE to tell you to remember null checks!!

ice1000 (approved):
bors r+

imkiva:
> > Suggest: replace all `@UnknownNullability` with `@Nullable` to force the IDE to tell you to remember null checks!!
> 
> Yes we can. Are you gonna do it?

I did part of it in this PR and there're only 2 `@UnknownNullability` left in `DefVar`. I guess it's fine now.

PR #311, opened 2021-12-22
Call `needTyck` with current module name

imkiva:
Check `needTyck` with current module name so imported definitions won't pollute `CompiledAya`

### DIFF
--- a/api/src/main/java/org/aya/api/ref/DefVar.java
+++ b/api/src/main/java/org/aya/api/ref/DefVar.java
@@ -54,4 +54,11 @@ public final class DefVar<Core extends CoreDef, Concrete extends ConcreteDecl> i
   @Override public boolean equals(Object o) {
     return this == o;
   }
+
+  public boolean isInModule(@NotNull ImmutableSeq<String> moduleName) {
+    var maybeSubmodule = module;
+    if (maybeSubmodule == null) return false;
+    if (maybeSubmodule.sizeLessThan(moduleName.size())) return false;
+    return maybeSubmodule.slice(0, moduleName.size()).equals(moduleName);
+  }
 }
--- a/api/src/main/java/org/aya/api/ref/DefVar.java
+++ b/api/src/main/java/org/aya/api/ref/DefVar.java
@@ -59,6 +59,6 @@ public final class DefVar<Core extends CoreDef, Concrete extends ConcreteDecl> i
     var maybeSubmodule = module;
     if (maybeSubmodule == null) return false;
     if (maybeSubmodule.sizeLessThan(moduleName.size())) return false;
-    return maybeSubmodule.slice(0, moduleName.size()).equals(moduleName);
+    return maybeSubmodule.sliceView(0, moduleName.size()).sameElements(moduleName);
   }
 }
--- a/base/src/main/java/org/aya/concrete/stmt/Decl.java
+++ b/base/src/main/java/org/aya/concrete/stmt/Decl.java
@@ -113,8 +113,8 @@ public sealed abstract class Decl extends Signatured implements Stmt, ConcreteDe
       this.ref = DefVar.concrete(this, name);
     }
 
-    @Override public boolean needTyck() {
-      return ref.concrete.signature == null;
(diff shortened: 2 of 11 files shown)


api/src/main/java/org/aya/api/ref/DefVar.java
```diff
@@ -54,4 +54,11 @@ private DefVar(Concrete concrete, Core core, @NotNull String name) {
   @Override public boolean equals(Object o) {
     return this == o;
   }
+
+  public boolean isInModule(@NotNull ImmutableSeq<String> moduleName) {
+    var maybeSubmodule = module;
+    if (maybeSubmodule == null) return false;
+    if (maybeSubmodule.sizeLessThan(moduleName.size())) return false;
```
ice1000, line 61: This line is probably useless

api/src/main/java/org/aya/api/ref/DefVar.java
```diff
@@ -54,4 +54,11 @@ private DefVar(Concrete concrete, Core core, @NotNull String name) {
   @Override public boolean equals(Object o) {
     return this == o;
   }
+
+  public boolean isInModule(@NotNull ImmutableSeq<String> moduleName) {
+    var maybeSubmodule = module;
```
ice1000, line 59: Does inlining this variable make any sense?

base/src/main/java/org/aya/concrete/stmt/Decl.java
```diff
@@ -113,8 +113,8 @@ public PrimDecl(
       this.ref = DefVar.concrete(this, name);
     }
 
-    @Override public boolean needTyck() {
-      return ref.concrete.signature == null;
+    @Override public boolean needTyck(@NotNull ImmutableSeq<String> currentMod) {
+      return ref.isInModule(currentMod) && ref.concrete.signature == null;
```
ice1000, line 117: Hmmm, so prims will be considered need-tyck in the current module. Is that desirable?

imkiva: > prims will be considered need-tyck in the current module 

because it may have concrete tele and result (see StmtTycker).

imkiva: And you may misunderstand the `currentModule`. It means the module who are using this `DefVar`

api/src/main/java/org/aya/api/ref/DefVar.java
imkiva: Just remind people that we should check submodules instead of simple `currentModule.equals(module)`

api/src/main/java/org/aya/api/ref/DefVar.java
imkiva: slice-and-compare is much slower than int comprisons

imkiva: And it's not useless as `.sliceView` may throw `IndexOutOfBounds`

ice1000: I didn't realize that kala sameElements didn't compare size 🤯 @Glavo

base/src/main/java/org/aya/concrete/stmt/Decl.java
ice1000: 👌🏼

api/src/main/java/org/aya/api/ref/DefVar.java
imkiva: > I didn't realize that kala sameElements didn't compare size 🤯 @Glavo

No. it's `.sliceView`, not `.sameElements`

ice1000:
Why is coverage decreasing this much?

ice1000:
image

😨

ice1000:
Ok, then I don't care.
bors r+

api/src/main/java/org/aya/api/ref/DefVar.java
imkiva: Just try `SeqView.of(1).sliceView(0, 10)`

ice1000 (approved):
lgtm

Issue #312, opened 2021-12-22
ClassCastException when a little change is made to the code in #11


refparo:
A minimized test:

```aya
public open import Paths

public open data Unit | unit
def coe {a b : Type} (eq : a = b) (x : a) (i : I) : eq.at i
  => arcoe eq.at x i
def idpath {A : Type} (a : A) : a = a => idp {A} {a}

-- *** Types

universe u

struct EventT : Type (lsuc u)
  | AgentT : Type u => Unit

def getAgentT (T : EventT) : Type u => T.AgentT
def setAgentT (A : Type u) (T : EventT) : EventT
  => new EventT { | AgentT => A }
def AgentT-inv (A : Type u) (T : EventT) : A = getAgentT (setAgentT A T)
  => idpath A


struct Event (T : EventT {universe u}) : Type (lsuc u)
  | agent : T.AgentT

def getAgent (T : EventT) (e : Event T) : T.AgentT => e.agent


universe v

def Quantifier (A : Type u) : Type (lmax u (lsuc v)) => (A -> Type v) -> Type v

def mkConstQ {A : Type u} (x : A) : Quantifier A
  => \f => f x


def Verb (T : EventT {universe u}) : Type (lmax (lsuc u) (lsuc v))
  => Quantifier (Event T) {universe (lsuc u), v}

def mkVerb {T : EventT} (p : Event T -> Type) : Verb T
  => \f => Sig (e : Event T) ** (Sig (p e) ** (f e))

def VP (T : EventT {universe u}) : Type (lmax (lsuc u) (lsuc v))
  => Quantifier (Event T) {universe (lsuc u), v}


def BNP (A : Type u) : Type (lmax u (lsuc v)) => Quantifier A {universe u, v}

def GNP (setT : Type u -> EventT -> EventT) (A : Type u)
  : Type (lmax (lsuc u) (lsuc v))
  => Pi {T : EventT} -> VP (setT A T) -> VP (setT A T)


def Theta (setT : Type u -> EventT -> EventT) : Type (lmax (lsuc u) (lsuc v))
  => Pi {A : Type u}
  -> BNP A {universe u, v}
  -> GNP setT A {universe u, v}

def mkTheta
  (getT : EventT -> Type u)
  (get : Pi (T : EventT) -> Event T -> getT T)
  (setT : Type u -> EventT -> EventT)
  (T-inv : Pi (A : Type u) (T : EventT) -> A = getT (setT A T))
  : Theta setT
[...]

PR #313, opened 2021-12-22
Remove reporter from library owner

imkiva:
When debugging #312 I found the reporter in library owner was not used and caused the compiler to skip writing core files for modules followed. It turns out some changes have been made to make the delayed reporter in the library compiler not working, which is mostly due to my careless reviewing of https://github.com/aya-prover/aya-dev/commit/b10d16ad26d5197836d992af378be0b4bdce2c71 which implicitly breaks the fixture in https://github.com/aya-prover/aya-dev/commit/c7b3ade253bff180ad2daf059e54f6c1a4e65add

hit hit kiva


ice1000 (approved):

ice1000:
Merge?

PR #322, opened 2022-02-11
[WIP] Resolve projection early

### LINKED ISSUE #214: Resolve projection early
So we can achieve easy mutual recursion on structs and functions that use these structs

imkiva:
This PR closes #214 

## Changes
- Struct fields are resolved in the resolver instead of the tycker.
- Structs can be `open`ed like `data`s.
- Fields can be projected by qualified names like `p.Paths::at`

# Problems
In [Cubical.aya](https://github.com/aya-prover/aya-dev/blob/7cd740cdc2a30c9de427fae7f3d155e6e22e7154/base/src/test/resources/success/common/src/Cubical.aya), opening all similar structs (like `ConnectedXXX`) causes name clashes when resolving the expr `c.p` in line 39.
We may save all possible fields in `ProjExpr` in resolver instead of reporting an error early and let the tycker find the correct field type-directedly.


ice1000:
> In [Cubical.aya](https://github.com/aya-prover/aya-dev/blob/7cd740cdc2a30c9de427fae7f3d155e6e22e7154/base/src/test/resources/success/common/src/Cubical.aya), opening all similar structs (like ConnectedXXX) causes name clashes when resolving the expr c.p in line 39.

I propose not opening those structs or rename the fields, because IMO the new behavior is the correct behavior.

Apart from that, maybe we need to address the resolution of `this` thing, like `new A { | fieldA => bla | fieldB => fieldA }`, in this case the latter `fieldA` reference should be desugared to `this.fieldA` or something. I conjecture that this is kinda difficult

imkiva:
> like `new A { | fieldA => bla | fieldB => fieldA }`, in this case the latter `fieldA` reference should be desugared to `this.fieldA` or something. I conjecture that this is kinda difficult

I guess we already handled this case in https://github.com/aya-prover/aya-dev/blob/31c8d4f6e7cbd8c537b1734c23d7eeae88eed6f0/base/src/main/java/org/aya/resolve/visitor/ExprResolver.java#L129-L131

ice1000:
> > like `new A { | fieldA => bla | fieldB => fieldA }`, in this case the latter `fieldA` reference should be desugared to `this.fieldA` or something. I conjecture that this is kinda difficult
> 
> I guess we already handled this case in
> 
> https://github.com/aya-prover/aya-dev/blob/31c8d4f6e7cbd8c537b1734c23d7eeae88eed6f0/base/src/main/java/org/aya/resolve/visitor/ExprResolver.java#L129-L131

Tql!

imkiva:
> > > like `new A { | fieldA => bla | fieldB => fieldA }`, in this case the latter `fieldA` reference should be desugared to `this.fieldA` or something. I conjecture that this is kinda difficult
> > 
> > 
> > I guess we already handled this case in
> > https://github.com/aya-prover/aya-dev/blob/31c8d4f6e7cbd8c537b1734c23d7eeae88eed6f0/base/src/main/java/org/aya/resolve/visitor/ExprResolver.java#L129-L131
> 
> Tql!

The code hits the breakpoint, but why

image

imkiva:
> Apart from that, maybe we need to address the resolution of `this` thing, like `new A { | fieldA => bla | fieldB => fieldA }`, in this case the latter `fieldA` reference should be desugared to `this.fieldA` or something. I conjecture that this is kinda difficult

Let's do `this` in another PR. It's complicated 😭

imkiva:
> > Apart from that, maybe we need to address the resolution of `this` thing, like `new A { | fieldA => bla | fieldB => fieldA }`, in this case the latter `fieldA` reference should be desugared to `this.fieldA` or something. I conjecture that this is kinda difficult
> 
> Let's do `this` in another PR. It's complicated 😭

I changed `IntroTerm.New`'s `@NotNull ImmutableMap<DefVar<FieldDef, Decl.StructField>, Term> params` to a `MutableMap` to reference `this` during struct construction and caused stack overflow in some visitors. 🤔

imkiva:
@ice1000 Yes, so we have to make our visitors stateful and I don't like this solution. We may introduce ThisTerm and keep the record in TyckState

imkiva:
I propose merging this PR early so we can test this commit https://github.com/aya-prover/aya-dev/commit/8afc7d17333d6c18e37deba61f929e2a6119dd4c

ice1000:
bors d=imkiva

imkiva:
bors merge

PR #323, opened 2022-02-16
Allow empty `new` block

### LINKED ISSUE #168: Allow `new Expr` without a block followed in parser
These should be empty records

imkiva:
This PR fixes #168 

### Details
Added a new rule of expr
```diff
 expr : atom                                 # single
      | expr argument+                       # app
      | NEW_KW expr LBRACE newArg* RBRACE    # new
+     | NEW_KW expr                          # newEmpty
```

rather than simply make the block part optional like this

```diff
 expr : atom                                 # single
      | expr argument+                       # app
-     | NEW_KW expr LBRACE newArg* RBRACE    # new
+     | NEW_KW expr (LBRACE newArg* RBRACE)? # new

```

This is because in the latter change for given code `new T {}` the antlr always matches the `{}` with the `expr` and treats the following as optional because there's a `{ exprList }` in `argument`. 

And we should allow both `new T` and `new T {}` for empty records for better error messages for the same reason (which is mostly due to antlr's lack of ability to pass parameters to subrules).

### DIFF
--- a/base/src/test/resources/success/src/Structs.aya
+++ b/base/src/test/resources/success/src/Structs.aya
@@ -18,3 +18,7 @@ struct Pos1 (A : Type) : Type
 struct Pos2 (B : Nat -> Type) : Type
   | x : Nat
   | y : B zero
+
+struct EmptyRecord : Type
+def new-empty1 => new EmptyRecord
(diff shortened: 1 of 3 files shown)


ice1000 (approved):
bors r+

PR #326, opened 2022-02-28
Mutual recursion on constructors and fields

imkiva:
TODOs:

+ [ ] Nuke `Sample`

This PR improves the inference of type checking order, including

- Header and body orders are generated at the same time, which is done by recording the part (header or body) where the resolver is working on.
- Constructors and fields are treated as top-level definitions during resolving and tyck order inference.
- Expr Resolver now has a hierarchical structure since we will treat ctors' and fields' dependencies as the body dependencies of the corresponding data and struct.


ice1000:
bors r=[ice1000, imkiva]

PR #331, opened 2022-03-14
Constructors and fields should explicitly depend on their top-level decls.

imkiva:
Fixed assertion error when compiling [Reflection.aya](https://github.com/aya-prover/aya-dev/blob/tactic/base/src/test/resources/success/common/src/Data/Reflection.aya)


imkiva:
bors merge

PR #338, opened 2022-03-21
Update the "semantic model" test with lower ulevels

refparo:


ice1000:
@refparo Good
bors r+

PR #356, opened 2022-03-27
Lazy `Term` visitor

wsx-ucb:
Cherry-picked a commit from `future` branch


base/src/main/java/org/aya/core/visitor/TermPostMap.java
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2020-2022 Yinsen (Tesla) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.core.visitor;
+
+import org.aya.core.term.Term;
+
+import java.util.function.Function;
+
+public record TermPostMap(TermView view, Function<Term, Term> f) implements TermView {
```
ice1000, line 9: Isn't this just `TermMap` but weaker?

base/src/main/java/org/aya/core/visitor/TermPreMap.java
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2020-2022 Yinsen (Tesla) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.core.visitor;
+
+import org.aya.core.term.Term;
+
+import java.util.function.Function;
+
+public record TermPreMap(TermView view, Function<Term, Term> f) implements TermView {
```
ice1000, line 9: Ditto

ice1000 (commented):
Remember to add not-null annotations and check CI.

I can confirm that the idea behind the code in `commit` overloads are mostly what I asked. I'll have to check more carefully to see if they're really correct though.

Great job!!

base/src/main/java/org/aya/core/visitor/TermPostMap.java
wsx-ucb: Yes, this is just `TermPostMap(view, id, f)`.
Not sure if we need this just for convenience.

wsx-ucb:
> I can confirm that the idea behind the code in commit overloads are mostly what I asked. I'll have to check more carefully to see if they're really correct though.

The `note/unfiled.md` has some information on the semantics.
The `traverse` function just follows the formalism there.
Also, the subtle conditions of composing pre/post functions are provided in the last paragraph.

ice1000:
> > I can confirm that the idea behind the code in commit overloads are mostly what I asked. I'll have to check more carefully to see if they're really correct though.
> 
> The `note/unfiled.md` has some information on the semantics. The `traverse` function just follows the formalism there. Also, the subtle conditions of composing pre/post functions are provided in the last paragraph.

Yeah!! The document is greeeaaaat

imkiva (approved):
Wow, the document is so good!!

base/src/main/java/org/aya/core/visitor/TermPostMap.java
wsx-ucb: I removed the `TermPreMap` and `TermPostMap` records, but still keep the methods.

ice1000: Great

base/src/main/java/org/aya/core/term/Term.java
```diff
@@ -216,4 +216,8 @@ public static boolean checkSubst(@NotNull SeqLike<@NotNull Param> params, @NotNu
     }
   }
 
+  default TermView view() {
+    var term = this;
```
ice1000, line 220: Why

wsx-ucb: Good catch, I guess we can inline that.

ice1000:
bros ping

ice1000:
bors ping

ice1000:
bros r+

Issue #358, opened 2022-03-29
Ghost story of REPL


ice1000:
![image](https://user-images.githubusercontent.com/16398479/160565049-78bf2e8a-56df-45e6-816a-90ac1e06c479.png)

![image](https://user-images.githubusercontent.com/16398479/160565221-0ac955a3-704f-4f14-83fb-cf029907a0db.png)

imkiva:
Ghost story 3: jar-ed REPL keeps throwing IllegalStateException, while the jlinked does not.

![image](https://user-images.githubusercontent.com/9902643/160624737-77ef238a-f530-409a-b5a1-3c070eff6fbb.png)

ice1000:
![image](https://user-images.githubusercontent.com/16398479/160769485-604b0636-6fff-4683-b65c-934abfba9b92.png)

ice1000:
I suspect it's caused by Java 18. Let's just assume that it is okay

PR #360, opened 2022-03-29
Use adaptive-cli style whenever outputting to a terminal

imkiva:
Fixes part of #358: the distiller colors changed after using `CliReporter`.

#### Changes 
- Do not make `UnixTermStylist` a singleton, instead, we need to tell it the selected color scheme.
- If we are outputting to a terminal (like REPL or cli compiler interface), always use adaptive color scheme ("adaptive" means use colors from user terminals' color scheme, rather than the hardcoded one)


ice1000:
Ok I see

ice1000:
bors d=imkiva

PR #369, opened 2022-04-06
REFACTOR: Implement full normalization with `TermView`.

wsx-ucb:
I've compared the new approach with the old one and there are two comments labeled `CHECK` that highlight the differences. May need to double check those changes are valid.

### DIFF
--- a/base/src/main/java/org/aya/core/visitor/TermOps.java
+++ b/base/src/main/java/org/aya/core/visitor/TermOps.java
@@ -169,7 +169,7 @@ public interface TermOps extends TermView {
         var result = PatMatcher.tryBuildSubstArgs(null, match.patterns(), args);
         if (result.isOk()) {
           subst.add(result.get());
-          var body = match.body().view().rename().subst(subst).normal(state).commit();
+          var body = match.body().view().subst(subst).normalize(state).commit();
           return new WithPos<>(match.sourcePos(), body);
         } else if (!orderIndependent && result.getErr())
           return null;
@@ -205,11 +205,10 @@ public interface TermOps extends TermView {
           if (def == null) yield fn;
           if (def.modifiers.contains(Modifier.Opaque)) yield fn;
           yield def.body.fold(
-            lamBody -> lamBody.view().subst(buildSubst(def.telescope(), fn.args())).rename().normal(state).commit(),
+            lamBody -> lamBody.view().subst(buildSubst(def.telescope(), fn.args())).normalize(state).commit(),
             patBody -> {
               var orderIndependent = def.modifiers.contains(Modifier.Overlap);
               var unfolded = unfoldClauses(orderIndependent, fn.args(), patBody);
-              // CHECK: Do we need to normalize `unfolded` again?
               return unfolded != null ? unfolded.data() : fn;
             }
           );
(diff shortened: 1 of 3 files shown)


base/src/main/java/org/aya/core/visitor/TermOps.java
```diff
@@ -132,4 +143,105 @@ public Term post(Term term) {
...
+          if (def == null) yield con;
+          var unfolded = unfoldClauses(true, con.conArgs(), def.clauses);
+          yield unfolded != null ? unfolded.data() : con;
+        }
+        case CallTerm.Fn fn -> {
+          var def = fn.ref().core;
+          if (def == null) yield fn;
+          if (def.modifiers.contains(Modifier.Opaque)) yield fn;
+          yield def.body.fold(
+            lamBody -> lamBody.view().subst(buildSubst(def.telescope(), fn.args())).rename().normal(state).commit(),
```
ice1000, line 208: Do we really need to `rename` here, given that our `subst` has capture-avoiding?

base/src/main/java/org/aya/core/visitor/TermOps.java
```diff
@@ -132,4 +143,105 @@ public Term post(Term term) {
...
+            patBody -> {
+              var orderIndependent = def.modifiers.contains(Modifier.Overlap);
+              var unfolded = unfoldClauses(orderIndependent, fn.args(), patBody);
+              // CHECK: Do we need to normalize `unfolded` again?
```
ice1000, line 212: Yes.

base/src/main/java/org/aya/core/visitor/TermOps.java
ice1000: (note that my comment also applies to other `rename` calls)

base/src/main/java/org/aya/core/visitor/TermOps.java
```diff
@@ -132,4 +143,105 @@ public Term post(Term term) {
...
+          } else {
+            var subst = buildSubst(fieldDef.fullTelescope(), access.args());
+            for (var field : fieldDef.structRef.core.fields) {
+              if (field == fieldDef) continue;
+              var fieldArgs = field.telescope().map(Term.Param::toArg);
+              var acc = new CallTerm.Access(access.of(), field.ref, access.structArgs(), fieldArgs);
+              subst.add(field.ref, IntroTerm.Lambda.make(field.telescope(), acc));
+            }
+            var unfolded = unfoldClauses(true, access.fieldArgs(), subst, fieldDef.clauses);
+            // CHECK: seems no need to substitute `unfolded` again with `subst`
```
ice1000, line 231: No.

base/src/main/java/org/aya/core/visitor/TermView.java
```diff
@@ -151,6 +152,10 @@ private Term traverse(Term term) {
     return subst.isEmpty() ? this : new TermOps.Subster(this, subst);
   }
 
+  default TermView normal(TyckState state) {
```
ice1000, line 155: `normalize`

ice1000 (commented):
Hope my comments are helpful :0

base/src/main/java/org/aya/core/visitor/TermOps.java
wsx-ucb: Ahh, the renaming is being done at substitution now, right?

base/src/main/java/org/aya/core/visitor/TermOps.java
wsx-ucb: I think everything unfolded are normalized before returning, see [here](https://github.com/aya-prover/aya-dev/pull/369/files#diff-349172ace414a6ad8b2b2a7fd8c4b9f90a8becf213add88455a7d42f886e535dR172).

ice1000: Gotcha

base/src/main/java/org/aya/core/visitor/TermOps.java
ice1000: Yaas

ice1000 (approved):
bors d=wsx-ucb

ice1000:
Whnfer when?

wsx-ucb:
I'll try to finish that in the following days. But we'll see if that needs to use the `TermView` API.

PR #370, opened 2022-04-07
TACTIC: phase 1 plan complete

lunalunaa:
- [x] Basic "Expr with hole" mechanism with stepwise refinement
- [x] Basic error reporting for tactics 
- [x] `ExprView` for expressions
- [x] Other minor improvements 
- [x] Isolate the tactic elaboration process

Currently, users can write code like [this](https://github.com/aya-prover/aya-dev/blob/062b2c09f8416aea33429f3388f8dbdbb41efa4a/base/src/test/resources/success/src/SimpleTac.aya), where users can create "tactic blocks" fill in holes in the "tactic head" one by one in the "tactic nodes" that follows, where these nodes can contain additional holes and be filled by nested nodes.

However, tactic blocks cannot be nested (see [this](https://github.com/aya-prover/aya-dev/blob/062b2c09f8416aea33429f3388f8dbdbb41efa4a/base/src/test/resources/failure/tactic/nested-tactic.aya)) because the holes are filled "in verbatim" from tactic node inputs. This means further tactic nodes will fill the first hole of the tactic head, and if this hole is contained in a tactic block it would be very problematic.

The result of elaborating tactic block is an expression that contains no holes (and no metas when elaborated into core term).

### DIFF
--- a/base/src/main/java/org/aya/concrete/visitor/ExprOps.java
+++ b/base/src/main/java/org/aya/concrete/visitor/ExprOps.java
@@ -3,6 +3,9 @@
 package org.aya.concrete.visitor;
 
 import org.aya.concrete.Expr;
+import org.aya.generic.util.InternalException;
+import org.aya.pretty.doc.Doc;
+import org.aya.util.error.SourcePos;
 import org.jetbrains.annotations.NotNull;
 
 public interface ExprOps extends ExprView {
@@ -17,5 +20,38 @@ public interface ExprOps extends ExprView {
 
   @Override default @NotNull Expr post(@NotNull Expr expr) {
     return view().post(expr);
+  }
+
+  class HoleFiller implements ExprOps {
+    final Expr placeHolder = new Expr.ErrorExpr(SourcePos.NONE, Doc.english("Internal Error for expr hole filler"));
+    boolean filled = false;
+    Expr filling = placeHolder;
+    Expr exprWithHole = placeHolder;
+
+    @Override public @NotNull ExprView view() {
+      return exprWithHole.view();
+    }
+
+    @Override public @NotNull Expr pre(@NotNull Expr expr) {
+      return switch (expr) {
+        case Expr.HoleExpr hole -> {
+          if (!filled) {
+            filled = true;
+            yield filling;
+          } else yield hole;
+        }
+        case Expr misc -> misc;
+      };
+    }
+
+    public @NotNull Expr fill(Expr exprWithHole, Expr filling) {
+      filled = false;
+      this.exprWithHole = exprWithHole;
+      this.filling = filling;
+      var filledExpr = commit();
+      if (!filled) throw new InternalException("Tactic elaborator cannot fill the hole of " + exprWithHole);
+
+      return filledExpr;
+    }
   }
 }
--- a/base/src/main/java/org/aya/concrete/visitor/ExprOps.java
+++ b/base/src/main/java/org/aya/concrete/visitor/ExprOps.java
@@ -49,7 +49,7 @@ public interface ExprOps extends ExprView {
       this.exprWithHole = exprWithHole;
       this.filling = filling;
       var filledExpr = commit();
-      if (!filled) throw new InternalException("Tactic elaborator cannot fill the hole of " + exprWithHole);
+      if (!filled) throw new InternalException("No hole has been found in " + exprWithHole);
 
       return filledExpr;
     }
--- a/base/src/main/java/org/aya/concrete/visitor/ExprView.java
+++ b/base/src/main/java/org/aya/concrete/visitor/ExprView.java
@@ -32,6 +32,17 @@ public interface ExprView {
     return new Expr.NamedArg(arg.explicit(), expr);
   }
 
+  private Expr.TacNode commit(Expr.TacNode node) {
+    return switch (node) {
+      case Expr.ExprTac exprTac -> commit(exprTac);
+      case Expr.ListExprTac listExprTac -> {
+        var tacNodes = listExprTac.tacNodes().map(this::commit);
+        if (tacNodes.sameElements(listExprTac.tacNodes(), true)) yield listExprTac;
+        yield new Expr.ListExprTac(listExprTac.sourcePos(), tacNodes);
+      }
+    };
+  }
+
   private @NotNull Expr traverse(@NotNull Expr expr) {
     return switch (expr) {
       case Expr.RefExpr ref -> ref; // I don't know
@@ -94,6 +105,11 @@ public interface ExprView {
       }
       case Expr.ErrorExpr error -> error;
       case Expr.MetaPat meta -> meta;
+      case Expr.TacExpr tac -> {
+        var node = commit(tac.tacNode());
+        if (node == tac.tacNode()) yield tac;
+        yield new Expr.TacExpr(tac.sourcePos(), node);
+      }
     };
   }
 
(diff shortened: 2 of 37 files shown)


lunalunaa:
try tactics but hole

base/src/main/java/org/aya/concrete/visitor/ExprFixpoint.java
```diff
@@ -118,4 +118,15 @@
...
+  default @NotNull Expr.TacNode visitTacNode(@NotNull Expr.TacNode node, P p) {
+    return switch (node) {
+      case Expr.TacNode.ExprTac expr -> new Expr.TacNode.ExprTac(expr.sourcePos(), expr.expr().accept(this, p));
+      case Expr.TacNode.ListExprTac list -> new Expr.TacNode.ListExprTac(list.sourcePos(), list.tacNodes().map(n -> visitTacNode(n, p)));
+    };
+  }
+
+  @Override default Expr visitTac(Expr.@NotNull TacExpr tactic, P p) {
+    return new Expr.TacExpr(tactic.sourcePos(), visitTacNode(tactic.tacNode(), p));
+  }
```
ice1000, line 131: Should cache

ice1000: Hmmm..

ice1000 (commented):
I think we should migrate every visitor to your exprview and nuke exprvisitors. I'll do this, stay tuned

base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -308,6 +310,23 @@ private Doc visitAccess(Stmt.@NotNull Accessibility accessibility, Stmt.Accessib
...
+  public @NotNull Doc tacNode(@NotNull Expr.TacNode tacNode) {
+    return switch (tacNode) {
+      case Expr.TacNode.ExprTac exprTac -> Doc.sep(Doc.symbol("|"), exprTac.expr().toDoc(options));
+      case Expr.TacNode.ListExprTac listExprTac -> tacList(listExprTac);
+    };
+  }
+
+  public @NotNull Doc tacList(@NotNull Expr.TacNode.ListExprTac tacList) {
+    return Doc.cat(
+      Doc.symbol("{"),
```
ice1000: `Doc.cblock`

ice1000: I'll do it

base/src/main/java/org/aya/concrete/visitor/ExprOps.java
```diff
@@ -0,0 +1,21 @@
...
+import org.aya.concrete.Expr;
+import org.jetbrains.annotations.NotNull;
+
+public interface ExprOps extends ExprView {
+  @NotNull ExprView view();
+  @Override default @NotNull Expr initial() {
+    return view().initial();
+  }
+
+  @Override default Expr pre(Expr expr) {
```
imkiva, line 14: It's better to note the parameter and return type with `@NotNull` or `@Nullable`

base/src/main/java/org/aya/concrete/visitor/ExprView.java
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * A generic view for traversing Expr
+ *
+ * @author Luna
+ */
+
+public interface ExprView {
+  @NotNull Expr initial();
+
+  default Expr pre(Expr expr) {return expr;}
```
imkiva, line 18: ditto

base/src/main/java/org/aya/core/term/Term.java
```diff
@@ -221,4 +222,29 @@ public static boolean checkSubst(@NotNull SeqLike<@NotNull Param> params, @NotNu
...
+
+  default @NotNull ImmutableSeq<Meta> allMetas() {
+    return new TermOps() {
+      final MutableList<Meta> metas = MutableList.create();
+
+      @Override public @NotNull TermView view() {
+        return Term.this.view();
+      }
+
+      @Override public Term pre(Term term) {
```
imkiva, line 234: `TermView` and `TermOps` should also be annotated with `@NotNull` or `@Nullable`, which is addressed in my HEAD. 😄

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -104,7 +105,8 @@ public AyaProducer(@NotNull Either<SourceFile, SourcePos> source, @NotNull Repor
     return unreachable(ctx);
   }
 
-  @NotNull private Remark visitRemark(AyaParser.RemarkContext remark) {
+  @NotNull
+  private Remark visitRemark(AyaParser.RemarkContext remark) {
     var sb = new StringBuilder();
```
imkiva, line 110:
```suggestion
  private @NotNull Remark visitRemark(AyaParser.RemarkContext remark) {
    var sb = new StringBuilder();
```

imkiva (approved):
LGTM with some nitpicks

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -338,10 +343,108 @@ else if (iter.hasNext()) {
...
     };
   }
 
+  private TacElabResult elaborateTactic(TacNode tacNode, Term term) {
+    return switch (tacNode) {
+      case TacNode.ExprTac exprTac -> new TacElabResult(exprTac.expr(), inherit(exprTac.expr(), term));
+      case TacNode.ListExprTac listExprTac -> {
+        var tacNodes = listExprTac.tacNodes();
+        var headNode = tacNodes.first();
+        var tailNodes = tacNodes.slice(1, tacNodes.size());
```
ice1000, line 372: Why not `drop(1)`?

lunalunaa: fixed

ice1000 (commented):
0. It would be good for the anonymous class in `allMetas` to be a record. Furthermore, you may use `VarConsumer` to replace it
1. You created `ExprFixpoint`. I guess it's due to the rebase. I think you should remove the file.

lunalunaa:
Addressing the comments:
0. Yes. I'll be changing it
1. This is probably a mistake when rebasing
2. `lastly` isn't really needed because one can override `commit`. But keeping it is also fine.

ice1000:
Niubi

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -366,19 +366,20 @@ else if (iter.hasNext()) {
...
     return switch (tacNode) {
       case TacNode.ExprTac exprTac -> new TacElabResult(exprTac.expr(), inherit(exprTac.expr(), term));
       case TacNode.ListExprTac listExprTac -> {
+        TacElabResult result = null;
+
         var tacNodes = listExprTac.tacNodes();
         var headNode = tacNodes.first();
         var tailNodes = tacNodes.drop(1);
+        // enter into a new local state
+        localCtx = localCtx.deriveMap();
```
ice1000, line 386: You can create a local of the old `localCtx` variable here, and use it later to avoid `requireNotNull`

ice1000 (commented):
Maybe you want to add some "usage of variables" test cases

base/src/main/java/org/aya/tyck/ExprTycker.java
lunalunaa: This is now fixed.

Yeah, there will a lot of more tests added tomorrow.

lunalunaa:
I will also be refactoring a bit because the function is really cluttered.

lunalunaa:
Closing this for now because I'm moving to v2.

ice1000:
@lunalunaa Maybe you still want to cherry pick some of the changes which might be useful for the future

PR #371, opened 2022-04-08
Move primitive related states to `TyckState`

imkiva:
## Changes
- `normalize` of Terms now accept a `@NotNull TyckState` (it is `@Nullable` before), because it holds primitives.
- `PrimDef.Factory` is no longer a singleton, instead, it is created every time the compilation task was created. (except for LSP, where we only clear the primitive factory)

## Tasks
- [x] Get rid of global states
- [x] Update tests
- [x] Test these changes in LSP, make sure it does not break any LSP feature

### DIFF
--- a/base/src/main/java/org/aya/concrete/remark/Literate.java
+++ b/base/src/main/java/org/aya/concrete/remark/Literate.java
@@ -122,7 +122,7 @@ public sealed interface Literate extends Docile {
 
     private @NotNull Doc normalize(@NotNull Term term) {
       var mode = options.mode();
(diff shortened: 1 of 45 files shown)


ice1000:
我靠

imkiva:
> 我靠

what's going on?

imkiva:
image

Yeah!! LSP w/ Library is still working!!!

ice1000:
tql. Will review in a day

ice1000 (approved):
bors r=dark-flames

Issue #381, opened 2022-04-12
`def a a` triggers NPE


lunalunaa:
As titled.

Should be something related to resolver.

Stacktrace here
```
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:208)
	at org.aya/org.aya.core.def.Def.defTele(Def.java:33)
	at org.aya/org.aya.tyck.ExprTycker.defCall(ExprTycker.java:466)
	at org.aya/org.aya.tyck.ExprTycker.inferRef(ExprTycker.java:437)
	at org.aya/org.aya.tyck.ExprTycker.doSynthesize(ExprTycker.java:67)
	at org.aya/org.aya.tyck.ExprTycker.synthesize(ExprTycker.java:397)
	at org.aya/org.aya.tyck.StmtTycker.lambda$checkTele$10(StmtTycker.java:299)
	at kala.collection/kala.collection.immutable.ImmutableVectors$Vector1.map(ImmutableVectors.java:371)
	at kala.collection/kala.collection.immutable.ImmutableVectors$Vector1.map(ImmutableVectors.java:162)
	at org.aya/org.aya.tyck.StmtTycker.checkTele(StmtTycker.java:296)
	at org.aya/org.aya.tyck.StmtTycker.tele(StmtTycker.java:284)
	at org.aya/org.aya.tyck.StmtTycker.tyckHeader(StmtTycker.java:136)
	at org.aya/org.aya.tyck.StmtTycker.doTyck(StmtTycker.java:65)
	at org.aya/org.aya.tyck.StmtTycker.traced(StmtTycker.java:54)
	at org.aya/org.aya.tyck.StmtTycker.tyck(StmtTycker.java:61)
	at org.aya/org.aya.tyck.order.AyaSccTycker.checkBody(AyaSccTycker.java:133)
	at org.aya/org.aya.tyck.order.AyaSccTycker.check(AyaSccTycker.java:117)
	at kala.collection/kala.collection.immutable.ImmutableVectors$Vector1.forEach(ImmutableVectors.java:440)
	at org.aya/org.aya.tyck.order.AyaSccTycker.checkMutual(AyaSccTycker.java:73)
	at org.aya/org.aya.tyck.order.AyaSccTycker.tyckSCC(AyaSccTycker.java:60)
	at org.aya.util/org.aya.util.tyck.OrgaTycker.tyckSCC(OrgaTycker.java:25)
[...]

PR #383, opened 2022-04-13
Fix many bugs

imkiva:
1. Fixed part of #381, now we can detect circular dependency like `def a a => xxxx`, but `def a a` is a more complicated problem, which will be addressed in another PR (possible solution #384).
2. Fixed library compiler's error reporting
3. The little typer should check the hole's solution, or the following code will crash the compiler
    ```
    def a => b
    def b => a
    ```
    when trying to solve metas generated for result type of `a` and `b`


ice1000:
With my commit, `def a : Type` will be reported with an error, but `def a a` still triggers NPE.

base/src/main/java/org/aya/tyck/StmtTycker.java
```diff
@@ -137,6 +138,8 @@ public void tyckHeader(@NotNull Decl decl, @NotNull ExprTycker tycker) {
         // It might contain unsolved holes, but that's acceptable.
         var resultRes = tycker.synthesize(fn.result).wellTyped().freezeHoles(tycker.state);
         fn.signature = new Def.Signature(resultTele, resultRes);
+        if (resultTele.isEmpty() && fn.body.isRight() && fn.body.getRightValue().isEmpty())
+          reporter.report(new NobodyError(decl.sourcePos, fn.ref));
```
imkiva, line 142: Can we do this check before tycking the telescope? So `def a a` would be rejected nicely.

ice1000: `def a a` does not have empty telescope.

ice1000: If your telescope uses yourself, you should reject the function

imkiva: Yes. My brain must be offline. 😢

ice1000:
Fixed conflicts

ice1000 (approved):
review submitted as commit
bors r+

PR #387, opened 2022-04-14
Keyword `I`

dark-flames:


base/src/main/java/org/aya/core/term/Term.java
```diff
@@ -137,6 +137,7 @@ default void traceExit(R r) {
     R visitFieldRef(@NotNull RefTerm.Field field, P p);
     R visitError(@NotNull ErrorTerm error, P p);
     R visitMetaPat(@NotNull RefTerm.MetaPat metaPat, P p);
+    R visitIntervalRef(@NotNull FormTerm.Interval interval, P p);
```
ice1000, line 140:
```suggestion
    R visitInterval(@NotNull FormTerm.Interval interval, P p);
```

base/src/main/java/org/aya/core/visitor/TermFixpoint.java
```diff
@@ -45,6 +45,11 @@ default int ulift() {
     return metaPat;
   }
 
+  @Override
+  @NotNull
+  default Term visitIntervalRef(FormTerm.@NotNull Interval interval, P p) {
```
ice1000, line 50:
```suggestion
  @Override default @NotNull Term visitIntervalRef(FormTerm.@NotNull Interval interval, P p) {
```

ice1000: And other similar things

ice1000 (changes requested):

base/src/main/java/org/aya/core/term/Term.java
dark-flames: I forgot to rename these methods after refactoring.

ice1000:
Conflicts

base/src/main/java/org/aya/core/pat/Pat.java
```diff
@@ -229,6 +229,64 @@ record Prim(
     }
   }
 
+  record Left(
+    boolean explicit
+  ) implements Pat {
```
ice1000, line 234:
```suggestion
  record Interval(
    boolean explicit,
    boolean isLeft
  ) implements Pat {
```

base/src/main/java/org/aya/core/pat/PatToTerm.java
```diff
@@ -30,6 +27,8 @@ public Term visit(@NotNull Pat pat) {
       case Pat.Bind bind -> new RefTerm(bind.bind(), 0);
       case Pat.Tuple tuple -> new IntroTerm.Tuple(tuple.pats().map(this::visit));
       case Pat.Meta meta -> new RefTerm.MetaPat(meta, 0);
+      case Pat.Left left -> new FormTerm.Left();
+      case Pat.Right right -> new FormTerm.Right();
```
ice1000, line 31: `Left` and `Right` are not form terms

base/src/main/java/org/aya/core/serde/SerPat.java
```diff
@@ -55,4 +55,22 @@ record Ctor(
       return new Pat.Ctor(explicit, state.resolve(name), params.map(param -> param.de(state)), ty.de(state));
     }
   }
+
+  record Left(
```
ice1000, line 59: Ditto

base/src/main/java/org/aya/tyck/pat/PatTycker.java
```diff
@@ -202,7 +194,18 @@ private void addPatSubst(@NotNull Var var, @NotNull Pat pat, @NotNull SourcePos
       }
       case Pattern.CalmFace face -> new Pat.Meta(face.explicit(), new Ref<>(),
         new LocalVar(Constants.ANONYMOUS_PREFIX, face.sourcePos()), term);
-      case default -> throw new UnsupportedOperationException("Number patterns are unsupported yet");
+      case Pattern.Number num -> {
```
ice1000, line 197: This is totally wrong. The type has to be interval to make `0`, `1` patterns.

base/src/main/java/org/aya/tyck/unify/DefEq.java
```diff
@@ -352,6 +352,8 @@ yield checkParam(lhs.param(), rhs.param(), null, lr, rl, () -> null, () -> {
         if (!compareLevel(lhs.lift(), rhs.lift())) yield null;
         yield new FormTerm.Univ((cmp == Ordering.Lt ? lhs : rhs).lift() + 1);
       }
+      case FormTerm.Interval lhs -> preRhs instanceof FormTerm.Interval rhs ? lhs : null;
+      case PrimTerm.End lhs -> preRhs instanceof PrimTerm.End rhs && lhs.val() == rhs.val() ? lhs : null;
```
ice1000, line 356: You return the type of both sides.

ice1000 (changes requested):

base/src/main/java/org/aya/tyck/pat/PatTycker.java
ice1000: Your code makes it impossible to have `0`, `1` patterns work for nat in the future

ice1000: Also image

ice1000 (commented):
Summary of a careful review:

0. The `org.aya.core.pat.Pat.End#toExpr` implementation is clearly wrong.
1. You always elaborate integer literals to interval endpoints, including both patterns and terms. What about `Nat`/`Int` literals?
2. Your understanding of `DefEq::doCompareUntyped` is wrong. You should look at existing code as implementation reference, like its handling of `ElimTerm.App`. It returns the type of both sides, but you returned the term. I don't think there is any reason to do this. Clearly, if two terms are equal, they are not the same (otherwise why not use `Objects.equals` for unification?), so you can't just return one of them. Clearly our test cases didn't even cover this! Probably we're implementing a **completely wrong** type theory from the very beginning without noticing..
3. There are some usages of `ImmutableMap` that I don't understand. I think it neither make sense for readability nor for maintainability.
4. I think my understanding of this PR is insufficient. I request an implementation summary, like what I did for #198.

ice1000 (commented):
More:

0. You should reformat your code. Please reformat your code. You may configure your IDE to reformat before a git commit.
1. You shouldn't use maps everywhere. Maps are expensive!! You can use arrays or functions.
2. You used integer bitwise or operation on booleans. I assume it's unintentional.

ice1000 (approved):
bors r+

ice1000:
> This PR was included in a batch that successfully built, but then failed to merge into main (it was a non-fast-forward update). It will be automatically retried.

??

PR #399, opened 2022-04-26
REPL enhancement

### LINKED ISSUE #382: NPE triggered because REPL context not cleaned up after tycking error
To trigger this bug in REPL, type `data Nat =`, followed by `Nat`.

The context is not properly cleaned up after catching exception. This causes `Nat` in the next line of input to be resolved into a `Def`.

https://github.com/aya-prover/aya-dev/blob/5e56d11bfde25b7fb3c0bb2cad592b24bbf1958c/cli/src/main/java/org/aya/cli/repl/ReplCompiler.java#L143-L145

### LINKED ISSUE #393: Parser error should not print stack trace in REPL
As titled

image

imkiva:
Fixes #382 fix #393

### DIFF
--- a/cli/src/main/java/org/aya/cli/repl/ReplCompiler.java
+++ b/cli/src/main/java/org/aya/cli/repl/ReplCompiler.java
@@ -2,10 +2,7 @@
 // Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
 package org.aya.cli.repl;
 
-import kala.collection.Seq;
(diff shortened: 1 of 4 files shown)


ice1000 (approved):
Maybe test the behavior?

ice1000 (approved):
bors d=imkiva

imkiva:
bors d=lunalunaa

ice1000:
IMO it's good enough. I would simply
bors r+

PR #400, opened 2022-04-28
Java integer backed literals

imkiva:
This PR implements literal numbers for Aya with the following features:
- Recognize representable datatypes like `Nat` with a flexible code shape matcher
- Meta-solving-based literal type decision for optimizing multiple datatypes at the same time.
- Pattern matching on literal

Future work:
- Recognize functions that operate on literal and replace them with Java backed operations (with pragma?)
- Polish the unification
- Tyck order for literal and data defs, the following code does not tyck
  ```
  def lit => 0
  data Nat | zero | suc Nat
  ```

For reviewers:
- Pay attention to `ExprTycker#doInherit`, not sure if the hole was generated in an ideal way.
- The `type` field of `LitTerm.ShapedInt` and `Pat.ShapedInt` looks annoying, but I failed to remove it :sob:

### DIFF
--- a/base/src/main/java/org/aya/core/pat/Pat.java
+++ b/base/src/main/java/org/aya/core/pat/Pat.java
@@ -227,6 +227,7 @@ public sealed interface Pat extends AyaDocile {
     }
   }
 
+  /** TODO[literal]: literal type needs meta-solving for first-class patterns */
   record ShapedInt(
     @Override int repr,
     @Override @NotNull AyaShape shape,
@@ -263,6 +264,10 @@ public sealed interface Pat extends AyaDocile {
 
     @Override public @NotNull Pat destruct(int repr) {
       return new Pat.ShapedInt(repr, this.shape, this.type, true);
+    }
+
+    @Override public @NotNull Pat self() {
+      return this;
     }
   }
 
--- a/base/src/main/java/org/aya/core/pat/Pat.java
+++ b/base/src/main/java/org/aya/core/pat/Pat.java
@@ -10,6 +10,7 @@ import org.aya.concrete.Expr;
 import org.aya.concrete.stmt.Decl;
 import org.aya.core.Matching;
 import org.aya.core.def.CtorDef;
+import org.aya.core.repr.AyaShape;
 import org.aya.core.term.CallTerm;
 import org.aya.core.term.RefTerm;
 import org.aya.core.term.Term;
@@ -222,6 +223,38 @@ public sealed interface Pat extends AyaDocile {
 
     @Override public void storeBindings(@NotNull LocalCtx localCtx) {
       // do nothing
+    }
+  }
+
+  record ShapedInt(
+    int integer,
+    @NotNull AyaShape shape,
+    // TODO: remove the type
+    @NotNull CallTerm.Data type,
+    boolean explicit
+  ) implements Pat {
+    @Override public @NotNull Expr toExpr(@NotNull SourcePos pos) {
+      return new Expr.LitIntExpr(pos, integer);
+    }
+
+    @Override public @NotNull Pat rename(@NotNull Subst subst, @NotNull LocalCtx localCtx, boolean explicit) {
+      return this;
+    }
+
+    @Override public @NotNull Pat zonk(@NotNull Tycker tycker) {
+      return this;
+    }
+
+    @Override public @NotNull Pat inline() {
+      return this;
+    }
+
+    @Override public void storeBindings(@NotNull LocalCtx localCtx) {
+      // do nothing
+    }
+
+    public @NotNull Pat constructorForm() {
+      return shape.transformPat(this, type);
     }
   }
 
--- a/base/src/main/java/org/aya/core/pat/Pat.java
+++ b/base/src/main/java/org/aya/core/pat/Pat.java
@@ -18,6 +18,7 @@ import org.aya.core.visitor.Subst;
 import org.aya.distill.BaseDistiller;
 import org.aya.distill.CoreDistiller;
 import org.aya.generic.Arg;
+import org.aya.generic.Shaped;
 import org.aya.generic.util.InternalException;
 import org.aya.pretty.doc.Doc;
 import org.aya.ref.DefVar;
@@ -227,14 +228,13 @@ public sealed interface Pat extends AyaDocile {
   }
 
   record ShapedInt(
-    int integer,
-    @NotNull AyaShape shape,
-    // TODO: remove the type
+    @Override int repr,
+    @Override @NotNull AyaShape shape,
     @NotNull CallTerm.Data type,
     boolean explicit
-  ) implements Pat {
+  ) implements Pat, Shaped.Inductively<Pat> {
     @Override public @NotNull Expr toExpr(@NotNull SourcePos pos) {
-      return new Expr.LitIntExpr(pos, integer);
+      return new Expr.LitIntExpr(pos, repr);
     }
 
     @Override public @NotNull Pat rename(@NotNull Subst subst, @NotNull LocalCtx localCtx, boolean explicit) {
@@ -253,8 +253,16 @@ public sealed interface Pat extends AyaDocile {
       // do nothing
     }
 
-    public @NotNull Pat constructorForm() {
-      return shape.transformPat(this, type);
+    @Override public @NotNull Pat makeZero(@NotNull CtorDef zero) {
+      return new Pat.Ctor(explicit, zero.ref, ImmutableSeq.empty(), type);
+    }
+
+    @Override public @NotNull Pat makeSuc(@NotNull CtorDef suc, @NotNull Pat pat) {
+      return new Pat.Ctor(explicit, suc.ref, ImmutableSeq.of(pat), type);
+    }
+
+    @Override public @NotNull Pat destruct(int repr) {
+      return new Pat.ShapedInt(repr, this.shape, this.type, true);
     }
   }
 
--- a/base/src/main/java/org/aya/core/pat/Pat.java
+++ b/base/src/main/java/org/aya/core/pat/Pat.java
@@ -15,6 +15,7 @@ import org.aya.core.term.CallTerm;
 import org.aya.core.term.RefTerm;
 import org.aya.core.term.Term;
 import org.aya.core.visitor.Subst;
+import org.aya.core.visitor.Zonker;
 import org.aya.distill.BaseDistiller;
 import org.aya.distill.CoreDistiller;
 import org.aya.generic.Arg;
@@ -243,7 +244,8 @@ public sealed interface Pat extends AyaDocile {
     }
 
     @Override public @NotNull Pat zonk(@NotNull Tycker tycker) {
-      return this;
+      // The cast must succeed
+      return new Pat.ShapedInt(repr, shape, (CallTerm.Data) tycker.zonk(type), explicit);
     }
 
     @Override public @NotNull Pat inline() {
--- a/base/src/main/java/org/aya/core/pat/Pat.java
+++ b/base/src/main/java/org/aya/core/pat/Pat.java
@@ -34,7 +34,7 @@ import org.jetbrains.annotations.NotNull;
 /**
  * @author kiva, ice1000
  */
-@Debug.Renderer(text = "toTerm().toDoc(DistillerOptions.DEBUG).debugRender()")
+@Debug.Renderer(text = "toTerm().toDoc(DistillerOptions.debug()).debugRender()")
 public sealed interface Pat extends AyaDocile {
   boolean explicit();
   default @NotNull Term toTerm() {
--- a/base/src/main/java/org/aya/core/pat/Pat.java
+++ b/base/src/main/java/org/aya/core/pat/Pat.java
@@ -15,7 +15,6 @@ import org.aya.core.term.CallTerm;
 import org.aya.core.term.RefTerm;
 import org.aya.core.term.Term;
 import org.aya.core.visitor.Subst;
-import org.aya.core.visitor.Zonker;
 import org.aya.distill.BaseDistiller;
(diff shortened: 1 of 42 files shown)


imkiva:
Needs test for
image

image

base/src/main/java/org/aya/core/pat/Pat.java
```diff
@@ -225,6 +227,50 @@ record End(boolean isRight, boolean explicit) implements Pat {
     }
   }
 
+  /** TODO[literal]: literal type needs meta-solving for first-class patterns */
```
ice1000, line 230: Can you explain this? I'm curious

ice1000:
I'm requesting a split of this PR. I wish to review this carefully (because I will need to use the APIs in the near future), but unfortunately a 1k+ LOC PR is too scary. I propose splitting it to:

+ Add shape matcher with `NAT_SHAPE`, maybe with tests
+ `ShapedInt` in core, with distill/ops/visitors (pain!! We should remove visitors completely so we no longer need all these burdens when adding new core terms)
+ Elaboration of shaped int with `Problem` and tests

imkiva:
@ice1000 👌

ice1000:
> image
> 
> It's half now!

Why is it this much less? I recall it's 1k+ loc and you extracted a 200 loc PR, how do these make this PR 500+ loc?

imkiva:
> > image
> > It's half now!
> 
> Why is it this much less? I recall it's 1k+ loc and you extracted a 200 loc PR, how do these make this PR 500+ loc?

See #404 

Now this PR only includes core changes, and others are included in #404

base/src/main/java/org/aya/core/pat/Pat.java
imkiva: If we have
```
pattern def pat => 0
```
We may solve the result type of `pat` regarding where `pat` is used.

ice1000: Meta should be generated per-use

base/src/main/java/org/aya/core/pat/Pat.java
```diff
@@ -225,6 +227,50 @@ record End(boolean isRight, boolean explicit) implements Pat {
...
+    @Override public @NotNull Expr toExpr(@NotNull SourcePos pos) {
+      return new Expr.LitIntExpr(pos, repr);
+    }
+
+    @Override public @NotNull Pat rename(@NotNull Subst subst, @NotNull LocalCtx localCtx, boolean explicit) {
+      return this;
+    }
+
+    @Override public @NotNull Pat zonk(@NotNull Tycker tycker) {
+      return this;
```
ice1000, line 246: need to zonk the data

base/src/main/java/org/aya/core/pat/Pat.java
```diff
@@ -225,6 +227,50 @@ record End(boolean isRight, boolean explicit) implements Pat {
...
+    @Override public @NotNull Pat inline() {
+      return this;
+    }
+
+    @Override public void storeBindings(@NotNull LocalCtx localCtx) {
+      // do nothing
+    }
+
+    @Override public @NotNull Pat makeZero(@NotNull CtorDef zero) {
+      return new Pat.Ctor(explicit, zero.ref, ImmutableSeq.empty(), type);
```
ice1000, line 258: Maybe make `type` a `Term` and report if `type` is not data call

base/src/main/java/org/aya/core/pat/PatUnify.java
```diff
@@ -29,18 +29,40 @@ private void unify(@NotNull Pat lhs, @NotNull Pat rhs) {
         else reportError(lhs, rhs);
       }
       case Pat.Ctor ctor -> {
-        if (rhs instanceof Pat.Ctor ctor1) {
-          // Assumption
-          assert ctor.ref() == ctor1.ref();
-          visitList(ctor.params(), ctor1.params());
-        } else reportError(lhs, rhs);
+        switch (rhs) {
```
ice1000, line 32: You can remove the braces outside of the switch

base/src/main/java/org/aya/core/pat/PatUnify.java
```diff
@@ -29,18 +29,40 @@ private void unify(@NotNull Pat lhs, @NotNull Pat rhs) {
...
+          }
+          case Pat.ShapedInt lit -> unifyLitWithCtor(ctor, lit, lhs);
+          default -> reportError(lhs, rhs);
+        }
       }
       case Pat.End end -> {
         if(!(rhs instanceof Pat.End rhsEnd)) reportError(lhs, rhs);
       }
+      case Pat.ShapedInt lhsInt -> {
+        switch (rhs) {
```
ice1000, line 46: Ditto

base/src/main/java/org/aya/core/repr/AyaShape.java
```diff
@@ -18,29 +18,40 @@
...
   }
 
-  record Factory(@NotNull MutableMap<Def, MutableList<AyaShape>> discovered) {
+  record Factory(@NotNull MutableMap<Def, AyaShape> discovered) {
     public Factory() {
       this(MutableLinkedHashMap.of());
     }
+
+    public @NotNull ImmutableSeq<Def> findImpl(@NotNull AyaShape shape) {
+      return discovered.view().map(Tuple::of)
```
ice1000, line 42: Is this `Tuple::of` necessary?

base/src/main/java/org/aya/core/repr/AyaShape.java
```diff
@@ -18,29 +18,40 @@
...
+        .map(t -> t._1)
+        .toImmutableSeq();
+    }
+
+    public @NotNull Option<AyaShape> find(@NotNull Def def) {
+      return discovered.getOption(def);
+    }
+
+    public void bonjour(@NotNull Def def, @NotNull AyaShape shape) {
+      // TODO[literal]: what if a def has multiple shapes?
```
ice1000, line 53: For now I believe there cannot be multiple shapes. I can only think of integers and lists for now.

base/src/main/java/org/aya/core/visitor/Unfolder.java
```diff
@@ -149,6 +150,10 @@
     return fieldBody.subst(arguments).accept(this, p);
   }
 
+  @Override @NotNull default Term visitShapedLit(@NotNull LitTerm.ShapedInt shaped, P p) {
+    return shaped.constructorForm();
```
ice1000, line 154: I don't think so

base/src/main/java/org/aya/generic/Shaped.java
```diff
@@ -0,0 +1,87 @@
...
+  @NotNull AyaShape shape();
+  @NotNull Term type();
+  @NotNull T constructorForm();
+  <O> boolean sameEncoding(@NotNull Shaped<O> other);
+
+  interface Inductively<T> extends Shaped<T> {
+    @Override @NotNull Term type();
+    @NotNull T makeZero(@NotNull CtorDef zero);
+    @NotNull T makeSuc(@NotNull CtorDef suc, @NotNull T t);
+    @NotNull T destruct(int repr);
```
ice1000, line 27: Did you mean `construct`?

base/src/main/java/org/aya/generic/Shaped.java
```diff
@@ -0,0 +1,87 @@
...
+
+    default @Override <O> boolean sameEncoding(@NotNull Shaped<O> other) {
+      if (shape() != other.shape()) return false;
+      if (!(other instanceof Inductively otherData)) return false;
+      var type = type();
+      var otherType = otherData.type();
+      return switch (type) {
+        case CallTerm.Data lhs && otherType instanceof CallTerm.Data rhs ->
+          lhs.ref().core == rhs.ref().core;
+        case CallTerm.Hole lhs && otherType instanceof CallTerm.Hole rhs ->
```
ice1000, line 39: Maybe `solveMeta`

ice1000: Idk

base/src/main/java/org/aya/generic/Shaped.java
```diff
@@ -0,0 +1,87 @@
...
+      return repr() == otherData.repr();
+    }
+
+    default @Override @NotNull T constructorForm() {
+      int repr = repr();
+      return with((zero, suc) -> {
+        if (repr == 0) return makeZero(zero);
+        return makeSuc(suc, destruct(repr - 1));
+      }, () -> {
+        // TODO[literal]: how to handle this?
```
ice1000, line 58: Report?

base/src/main/java/org/aya/generic/Shaped.java
```diff
@@ -0,0 +1,87 @@
...
+      if (!sameEncoding(other)) return false;
+      var otherData = ((Inductively<O>) other);
+      return repr() == otherData.repr();
+    }
+
+    default @Override @NotNull T constructorForm() {
+      int repr = repr();
+      return with((zero, suc) -> {
+        if (repr == 0) return makeZero(zero);
+        return makeSuc(suc, destruct(repr - 1));
```
ice1000, line 56: Loops instead of recursion?

base/src/main/java/org/aya/generic/Shaped.java
```diff
@@ -0,0 +1,87 @@
...
+      }, () -> {
+        // TODO[literal]: how to handle this?
+        throw new InternalException("trying to make constructor form without type solved");
+      });
+    }
+
+    default <R> R with(
```
ice1000, line 63: Do we have access to `TyckState`? If so, we should check solvedness using `TyckState`

ice1000 (commented):
This is incomplete

base/src/main/java/org/aya/generic/Shaped.java
imkiva: There's no recursion here.

base/src/main/java/org/aya/core/repr/AyaShape.java
imkiva: Yes. The MapView didn't have some functions like filter, etc.

base/src/main/java/org/aya/core/pat/Pat.java
imkiva: Yes I did this before but I cannot come up with a case that the pattern's type is a hole. Maybe we should keep the code consistent between literal Terms and Pats.

base/src/main/java/org/aya/core/visitor/Unfolder.java
imkiva: I forgot to visit its type, is it?

base/src/main/java/org/aya/generic/Shaped.java
imkiva: It means "when this function is called, current literal is destructed". So I think the name is fine.

base/src/main/java/org/aya/core/repr/AyaShape.java
imkiva: Then let's be a TODO.

base/src/main/java/org/aya/core/visitor/Unfolder.java
imkiva: No I don't think so

base/src/main/java/org/aya/generic/Shaped.java
imkiva: I don't know, but this should not be user related I think. We may have some methods to continue with unresolved holes?

At now, only calls from `DefEq` may trigger this error, do you think it's good to check the solveness before calling this function there?

base/src/main/java/org/aya/core/pat/Pat.java
ice1000: Well Idk either, will `def f (a b : _) : a + b = b + a` work? But it seems that the types will be solved before checking the patterns. Hmmm...

base/src/main/java/org/aya/core/repr/AyaShape.java
ice1000: Geez

ice1000:
> I forgot to visit its type, is it?

I believe that we should not unfold literals to their constructor forms. I think they should be kept as-is.

base/src/main/java/org/aya/generic/Shaped.java
ice1000: > I don't know, but this should not be user related I think. We may have some methods to continue with unresolved holes?
> 
> At now, only calls from `DefEq` may trigger this error, do you think it's good to check the solveness before calling this function there?

Maybe we can leave this TODO and see if this can actually be triggered.

base/src/main/java/org/aya/core/pat/Pat.java
imkiva: > Well Idk either, will `def f (a b : _) : a + b = b + a` work?

Currently this code does not tyck because the hole was unsolved (my test case follows)
```
def f (a b : _) : Nat
  | 0, 0 => 0
  | a, b => a
```

> But it seems that the types will be solved before checking the patterns

Yes. But if the types are unsolved, no patterns will be checked further.

imkiva: If we make the `type` of Pat a `Term`, we may add some yet-unreachable code. So I propose to do this when we need to solve the type of patterns.

base/src/main/java/org/aya/generic/Shaped.java
imkiva: Now we will check solutions for terms, see https://github.com/aya-prover/aya-dev/pull/400/commits/6f576d6ca9912bd921f80bfabc0b9e89bcdebd1e

ice1000 (approved):
bors r+

ice1000:
400!

Issue #51, opened 2021-02-09
Pragma design


ice1000:
How do we design pragmas? This is needed for the design of prelude.

lunalunaa:
> + :tada: `@pragma(parameters)`
> 
> + :eyes: `#[inline(parameters)]`
> 
> + :rocket: `%inline parameters`
> 
> + :smile: special comments

~~Breaking the tie~~

ice1000:
> > + :tada: `@pragma(parameters)`
> > 
> > + :eyes: `#[inline(parameters)]`
> > 
> > + :rocket: `%inline parameters`
> > 
> > + :smile: special comments
> 
> ~~Breaking the tie~~

Lmfao 🤣🤣🤣🤣

PR #403, opened 2022-04-29
Parse time desugaring of do notations, array syntax and list comprehension, and idiom brackets

### LINKED ISSUE #396: Support monad syntax
[Agda-style](https://agda.readthedocs.io/en/v2.6.2.1/language/syntactic-sugar.html) syntactic sugar should suffice for the purpose of tactic monad

lunalunaa:
## New features:

- Closes #396 : Do-notation that is desugared immediately at parse time.
- Array, also desugared immediately at parse time.
- Idiom bracket in the style of Agda, also desugared immediately at parse time.
- list comprehension, also at parse time

## Caveats

- Desugaring into whatever relevant operators in scope.
- Associativity is implementation-defined.

## Future work
- #51 : Pragmas could be used to support arbitrary resolving.
- Feel free to suggest more


base/src/main/java/org/aya/ref/LocalVar.java
```diff
@@ -7,10 +7,17 @@
...
 
+import java.util.UUID;
+
 /**
  * @author ice1000
  */
 public record LocalVar(@NotNull String name, @NotNull SourcePos definition) implements Var {
+
+  public static LocalVar IGNORED_LOCAL() {
+    return new LocalVar("AYA_INTERNAL_IGNORED_LOCALVAR" + UUID.randomUUID());
```
ice1000, line 18: If it's ignored, is it never substituted? Can we make it a constant instead?

base/src/test/resources/success/common/src/Data/List.aya
```diff
@@ -7,74 +7,114 @@ public open data List (A : Type) : Type
 | infixr :< A (List A)
 bind tighter =
 
-variable A B C : Type
```
ice1000, line 10: Why delete this?

buildSrc/src/main/antlr/org/aya/parser/AyaLexer.g4
```diff
@@ -49,14 +49,17 @@ EXTENDS : 'extends';
...
 
 // Unimplemented but reserved
-DO_KW : 'do';
 CODATA_KW : 'codata';
 LET_KW : 'let';
 IN_KW : 'in';
 COMPLETED : 'completed';
 
+// reserved
+IGNORED_LOCAL : 'AYA_INTERNAL_IGNORED_LOCALVAR';
```
ice1000, line 61: How is this useful

ice1000 (commented):
Maybe also refractor `a → b` with "ignored param"

base/src/main/java/org/aya/ref/LocalVar.java
lunalunaa: I'm not sure non-unique names in lambda arguments are acceptable, if so then constants probably wouldn't work.

base/src/test/resources/success/common/src/Data/List.aya
lunalunaa: See code below this line, I have to use metas that depends on other metas.

buildSrc/src/main/antlr/org/aya/parser/AyaLexer.g4
lunalunaa: Indeed this is not useful considering I generated unique names.

base/src/test/resources/success/src/Assoc.aya
```diff
@@ -71,10 +71,10 @@ def infixl |=> (a b : Dir) : Sig Dir ** Dir => (a, b)
 
 def id {A : Type} (a : A) : A => a
 
-def secR => (|=> id id id id R)
+def secR => ( |=> id id id id R)
 def secL => (id L |=>)
 
-def secR' => (|=> R)
+def secR' => ( |=> R)
 def secL' => (L |=>)
```
imkiva, line 78: The reformat seems unnecessary since spaces are not inserted after `|=>`.

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -366,11 +366,94 @@ public QualifiedID visitQualifiedId(AyaParser.QualifiedIdContext ctx) {
...
         visitExpr(forall.expr()));
+      case AyaParser.DoContext doCtx -> visitDo(doCtx);
+      case AyaParser.IdiomContext idmCtx -> {
+        if (idmCtx.idiomBlock().barredExpr() == null)
+          yield new Expr.UnresolvedExpr(sourcePosOf(idmCtx), "empty");
+        yield visitIdiomBlock(idmCtx.idiomBlock());
+      }
+      case AyaParser.ArrayContext arrCtx -> {
+        if (arrCtx.exprList() == null)
+          yield new Expr.UnresolvedExpr(sourcePosOf(arrCtx), "nil");
```
imkiva, line 377: Consider extracting constant strings to `Constants` class

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -366,11 +366,94 @@ public QualifiedID visitQualifiedId(AyaParser.QualifiedIdContext ctx) {
...
+        yield visitArr(arrCtx);
+      }
       // TODO: match
       default -> throw new UnsupportedOperationException("TODO: " + ctx.getClass());
     };
   }
 
+  private @NotNull Expr visitArr(AyaParser.ArrayContext ctx) {
+    var consArg = new Expr.NamedArg(true, new Expr.UnresolvedExpr(sourcePosOf(ctx), ":<"));
+    var nilArg = new Expr.NamedArg(true, new Expr.UnresolvedExpr(sourcePosOf(ctx), "nil"));
```
imkiva, line 387: ditto

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -366,11 +366,94 @@ public QualifiedID visitQualifiedId(AyaParser.QualifiedIdContext ctx) {
...
+      .appended(nilArg)
+      .toImmutableSeq();
+    return new Expr.BinOpSeq(sourcePosOf(ctx), args);
+  }
+
+  /**
+   * Warning: the parser cannot enforce left associativity at this stage
+   */
+  private @NotNull Expr visitIdiomBlock(AyaParser.IdiomBlockContext ctx) {
+    var orArg = new Expr.NamedArg(true, new Expr.UnresolvedExpr(sourcePosOf(ctx), "<*>"));
```
imkiva, line 400: Does this mean users should declare the `infixl` for `<*>`?

imkiva: But indeed you can enforce left-assoc by putting binary args in a `newBinOPScope`, for example, the code `a <*> b <*> c` can be parsed as `newBinOPScope(newBinOPScope(a, <*>, b), <*>, c)` which will ignore the assiciativity of `<*>` since there's only two operands.

base/src/test/resources/success/common/src/Data/List.aya
imkiva: Will keeping existing code generalized while using hand-inserted param for newly added code work?

buildSrc/src/main/antlr/org/aya/parser/AyaLexer.g4
lunalunaa: Actually no. There is a possibility that this might end up in the distiller and we want to report a '_' based on the prefix of the var name. So this is still better left internal (with some slight changes).

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
lunalunaa: Will do.

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
lunalunaa: Will do.

base/src/test/resources/success/src/Assoc.aya
imkiva: Ohhhh. Then we should replace the test case with a better name IMO

base/src/main/java/org/aya/ref/LocalVar.java
lunalunaa: Will address this in a future PR

ice1000: This is a special name anyway.

buildSrc/src/main/antlr/org/aya/parser/AyaLexer.g4
ice1000: This does not make `AYA_INTERNAL_IGNORED_LOCALVAR_123` reserved.

ice1000: Sorry I meant `AYA_INTERNAL_IGNORED_LOCALVAR123`

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
ice1000: I prefer local open btw

base/src/test/resources/success/src/Assoc.aya
ice1000: Oh no

ice1000 (changes requested):
Two minor comments, rest lgtm:

+ I still don't like the IGNORED LOCAL VAR thing. I think this is an abuse of the parser. I propose introducing a global constant for unused var and add special case for it in the substitutor and defeq (for the `lr`, `rl` maps). If you're unsure about this, I can implement this and you base your work on the 'ignored var' implemented by me later.
+ `variable A B C : Type` is not added back

lunalunaa:
> Two minor comments, rest lgtm:
> 
> 
> 
> + I still don't like the IGNORED LOCAL VAR thing. I think this is an abuse of the parser. I propose introducing a global constant for unused var and add special case for it in the substitutor and defeq (for the `lr`, `rl` maps). If you're unsure about this, I can implement this and you base your work on the 'ignored var' implemented by me later.
> 
> + `variable A B C : Type` is not added back

I agree. Looking forward to your implementation.

imkiva:
I am going to push some commits here

imkiva:
@ice1000 @lunalunaa Please review this commit (add support for ignored var): https://github.com/aya-prover/aya-dev/pull/403/commits/dd6f1651aa9e214f8a29cb05baf6d22025699058

imkiva:
> I'm not around my computers but @ice1000 is currently enhancing this to skip some normalizations.

It's been image, and I am sure @ice1000 still need that commit (it is mainly about making ignored vars unreferenceable)

ice1000:
Several leftovers:

+ I still wish idiom brackets to be desugared in the desugarer, so `(| 1 + 2 |)` would work.
+ It would be nice if `LocalVar.IGNORED` are generated for unused telescopes, but doing so fails the assertion. I think this might mean that we have substitution bugs (maybe we need to formalize everything like Idris2 did...).
+ There isn't something like `do(myBind, myReturn)`, and we don't have local openings. I guess we may want local openings.

ice1000 (approved):
bors r+

Issue #386, opened 2022-04-13
Overhaul struct


mio-19:
This means that we will support [arend-style anonymous-extensions](https://arend-lang.github.io/about/arend-features#anonymous-extensions) and `extends` as well as `this` expression.

Structs have arguments no more and we will use partially instantiation to express thing that are currently expressed using arguments, which means that following 2 definitions are equivalent.
```
open struct Pair (A : Type) (B : Type) : Type
  | fst : A
  | snd : B
  | we-are-together : Sig A ** B => (fst, snd)
```

```
open struct Pair : Type
  | A : Type
  | B : Type
  | fst : A
  | snd : B
  | we-are-together : Sig A ** B => (fst, snd)
```
However, we might want to remove `fields` instead of `telescope` of the `StructDef` class in the current implementation of aya for the reason that `telescope` is defined in a superclass of `StructDef` called Def.

For the `this` expression, we will support the keyword `this` and `this@context` for nested struct expressions.

imkiva:
Does the code in `classable` branch mean _**we only accept the second definition?**_

tsao-chi:
@imkiva The first definition is accepted as a syntactic sugar

imkiva:
> @imkiva The first definition is accepted as a syntactic sugar

I don't think we need an extra `ClassDecl` in this case, we can delay the desugar of telescope in the tycker and produce a tele-free `StructDef`.

imkiva:
> > @imkiva The first definition is accepted as a syntactic sugar
> 
> I don't think we need an extra `ClassDecl` in this case, we can delay the desugar of telescope until we are tycking them and produce a tele-free `StructDef`.

So we can get rid of these annoying and complicated inheritance relationships

tsao-chi:
@imkiva We can reverted to previous inheritances

imkiva:
> @imkiva We can reverted to previous inheritances

I tried some refactorings but it became worse because we have made lots of assumptions of concrete decls (like the `Signatured`, the `result` and the `ctx`). I think we just need some simplification based on your work (no need to revert)

imkiva:
> I tried some refactorings but it became worse because we have made lots of assumptions of concrete decls (like the `Signatured`, the `result` and the `ctx`). I think we just need some simplification based on your work (no need to revert)

see #428

tsao-chi:
Maybe we should consider more fully on taking this change.
This overhaul brings a lot more 工程方面的 difficulties.
[...]

PR #406, opened 2022-05-04
add api for classable def

tsao-chi:
This is a part of https://github.com/aya-prover/aya-dev/issues/386

Currently aya only has callable def, which is definition with some linear anonymous arguments. However #386 will introduce a new type of def, which is called classable def. (`class = struct`)

This PR contains minimal changes and does introduce inconsistency in naming. For example, `def` might refer to callable def or all def. This inconsistency will be handled by future refactors.


base/src/main/java/org/aya/core/def/GenericDef.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.aya.concrete.stmt.Decl;
+import org.aya.ref.DefVar;
+import org.aya.ref.GenericDefVar;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * @author zaoqi
+ */
+public sealed interface GenericDef permits ClassDef, Def {
```
ice1000, line 13: Maybe more methods? Like `accessibility`?

ice1000 (changes requested):
Can you merge `GenericDefVar`, `ClassDefVar` into `DefVar`?

Rest lgtm

ice1000:
@tsao-chi please address comments (I believe it's easy), I wish to merge this fast and furiously

base/src/main/java/org/aya/concrete/stmt/Stmt.java
```diff
@@ -23,7 +23,7 @@
  * @author kiva
  */
 public sealed interface Stmt extends AyaDocile, TyckUnit
-  permits Command, Decl, Generalize, Remark {
+        permits Remark, ClassDecl, Command, Decl, Generalize {
```
imkiva, line 26: This indent is not needed because other `permits` does not have leading spaces.

base/src/main/java/org/aya/distill/CoreDistiller.java
```diff
@@ -127,7 +127,8 @@ yield visitCalls(false, term(Outer.AppHead, head), args.view(), outer,
       }
       case CallTerm.Prim prim -> visitArgsCalls(prim.ref(), FN_CALL, prim.args(), outer);
       case RefTerm.Field term -> linkRef(term.ref(), FIELD_CALL);
-      case ElimTerm.Proj term -> Doc.cat(term(Outer.ProjHead, term.of()), Doc.symbol("."), Doc.plain(String.valueOf(term.ix())));
+      case ElimTerm.Proj term ->
+        Doc.cat(term(Outer.ProjHead, term.of()), Doc.symbol("."), Doc.plain(String.valueOf(term.ix())));
```
imkiva, line 131: Don't reformat existing code unless it is really unformatted.

base/src/main/java/org/aya/distill/CoreDistiller.java
```diff
@@ -191,12 +192,13 @@ private ImmutableSeq<Arg<Term>> visibleArgsOf(CallTerm call) {
       case Pat.Absurd absurd -> Doc.bracedUnless(Doc.styled(KEYWORD, "()"), absurd.explicit());
       case Pat.Tuple tuple -> Doc.licit(tuple.explicit(),
         Doc.commaList(tuple.pats().view().map(sub -> pat(sub, Outer.Free))));
-      case Pat.End end -> Doc.bracedUnless(Doc.styled(KEYWORD, !end.isRight() ? "0": "1"), end.explicit());
+      case Pat.End end -> Doc.bracedUnless(Doc.styled(KEYWORD, !end.isRight() ? "0" : "1"), end.explicit());
```
imkiva, line 195: This code is unformatted, so this change is OK.

tsao-chi:
> Can you merge `GenericDefVar`, `ClassDefVar` into `DefVar`?
> 
> Rest lgtm

It appears to be not that easy I previously thought

base/src/main/java/org/aya/core/def/GenericDef.java
tsao-chi: Def doesn't have `accessibility`

ice1000:
> > Can you merge `GenericDefVar`, `ClassDefVar` into `DefVar`?
> > 
> > Rest lgtm
> 
> It appears to be not that easy I previously thought

Can you elaborate? I am curious. What usage of DefVar/GenericVar makes it uneasy to use?

tsao-chi:
> > > Can you merge `GenericDefVar`, `ClassDefVar` into `DefVar`?
> > > Rest lgtm
> > 
> > 
> > It appears to be not that easy I previously thought
> 
> Can you elaborate? I am curious. What usage of DefVar/GenericVar makes it uneasy to use?

There were a number of places where `DefVar` is assumed to have a def with telescope such as `resolveStmt`

ice1000:
@tsao-chi You can use if (defvar.concrete instanceof), I think it's okay

ice1000 (approved):
bors r+

PR #409, opened 2022-05-10
Decl no visitor! Stay at home!

ice1000:
This PR makes `StmtOps` a pattern-matcher and deletes all concrete decl visitors.


ice1000:
@tsao-chi your work should be greatly simplified with this PR

PR #410, opened 2022-05-10
Delete expr visitors, release 0.18 (for guest0x0 use)

ice1000:
bors r+


base/src/main/java/org/aya/concrete/Expr.java
```diff
@@ -35,15 +35,6 @@
...
 public sealed interface Expr extends AyaDocile, SourceNode {
-  <P, R> R doAccept(@NotNull Visitor<P, R> visitor, P p);
-
-  default <P, R> R accept(@NotNull Visitor<P, R> visitor, P p) {
-    visitor.traceEntrance(this, p);
-    var ret = doAccept(visitor, p);
-    visitor.traceExit(ret, this, p);
-    return ret;
-  }
-
```
imkiva, line 46: Will this change break the tyck tracer?

imkiva: Okay, I found the `tracing` method in `ExprTycker`

PR #412, opened 2022-05-16
String literal with primitive concatenation

lunalunaa:
As titled.


base/src/main/java/org/aya/core/def/PrimDef.java
```diff
@@ -208,7 +246,10 @@ public enum ID {
     /** Short for <em>Arend coe</em>. */
     ARCOE("arcoe"),
     SQUEEZE_LEFT("squeezeL"),
-    INVOL("invol");
+    INVOL("invol"),
+    STR("Str"),
```
imkiva, line 250: Suggestion: change the type name to `String`. `Str` is just so ugly

imkiva (commented):
LGTM

base/src/main/java/org/aya/core/def/PrimDef.java
ice1000: I agree, you can always use a renaming open to use a shorter name locally

base/src/main/java/org/aya/tyck/LittleTyper.java
```diff
@@ -77,6 +78,7 @@ public record LittleTyper(@NotNull TyckState state, @NotNull LocalCtx localCtx)
       case FormTerm.Univ univ -> new FormTerm.Univ(univ.lift() + 1);
       case FormTerm.Interval interval -> new FormTerm.Univ(0);
       case PrimTerm.End end -> FormTerm.Interval.INSTANCE;
+      case PrimTerm.Str str -> state.primFactory().getOption(PrimDef.ID.STR).get().result();
```
ice1000, line 81: Are you sure it's `.get().result()` not `getCall(ID.STR)`?

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -270,7 +270,8 @@ public QualifiedID visitQualifiedId(AyaParser.QualifiedIdContext ctx) {
       throw new ParsingInterruptedException();
     }
     var string = ctx.STRING();
-    if (string != null) return new Expr.LitStringExpr(pos, string.getText());
+    if (string != null)
+      return new Expr.LitStringExpr(pos, string.getText().substring(1, string.getText().length() - 1));
```
ice1000, line 274: Escape

base/src/main/java/org/aya/tyck/LittleTyper.java
lunalunaa: No it's just I forgot.

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
lunalunaa: I'm not sure what this means. Should I do `translateEscapes()` here?

ice1000: image

PR #420, opened 2022-05-23
Optimize `PatClassifier` for literals

imkiva:
We used to convert literals to constructor forms when classifying them, so it's possible to overflow the stack if the literal was too large. 

But when classifying literal patterns together with other patterns, literal patterns will be converted to constructor forms whenever there are remaining constructor patterns, for example, when calling `classify([20, suc a])`, the patterns below
```
20 => 
suc a =>
```
will be treated as
```
suc 19 =>
suc a =>
```

So we put them in an `MCT.Node` and remove their heads to form the next-step-patterns `[19, a]` and recursively call `classify([19, a])`, which may cause StackOverflow if the literal 19 becomes larger.

It's easy to find the optimizable case: there always exists a scenario that **the patterns to be classified in the next step has only two kinds of patterns: bind and literal**, like `[19, a]` or `[19, 19, 20, a]`.

This PR takes into account the situation described above and replaces the convert-then-match method with a  simple `groupingBy` to classify literals when possible.

note: this PR contains 502 addition, but don't panic because most of them are tests.

Sorry for my English

### DIFF
--- a/base/src/main/java/org/aya/tyck/pat/PatClassifier.java
+++ b/base/src/main/java/org/aya/tyck/pat/PatClassifier.java
@@ -31,6 +31,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.function.BiFunction;
+import java.util.stream.Collectors;
 
 /**
  * @author ice1000, kiva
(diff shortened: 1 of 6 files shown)


ice1000 (approved):
bors r+

Apologies for the delay.

PR #427, opened 2022-06-25
General `Term` traversal

wsx-ucb:
Providing a new API to construct a global map `folded : Term -> R` (in the sense of obtaining `R` from the entirety of `Term`) by providing a local map `fold : Tm<R> -> R` (in the sense of only merging the results from the immediate sub-`Term`s).
Dually, there is a way to construct `unfolded : R -> Term` from a map `unfold : R -> Tm<R>`.


imkiva (changes requested):
Please make use of nullability annotations in those visitors

wsx-ucb:
I can definitely add those annotations in a later commit. I'm just wondering if there is a general guideline for annotations? It seems that 90% of the time we want `@NotNull`, and for the few exceptions we use `@Nullable` to indicate some special states. Is it possible to make `@NotNull` the default? Otherwise excessive annotations like `@NotNull ImmutableSeq<@NotNull Arg<@NotNull Term>>>` make function signatures quite unreadable IMO. Also, what about other annotations such as `@Override` and `@Contract`? (Sorry for my n00b questions)

ice1000:
@wsx-ucb Those are basically added by "hmm, people might think this is nullable, let's add notnull" criteria. Sometimes if I deal with the null case of a nullable value, then I add notnull to everywhere that uses this value.
Override is always added.
Contract is added when you think it can assist the dataflow analysis.

ice1000:
So, for example, I usually don't add notnull to immseq because we hardly use a null value on them. However, TyckState/Term may worth a notnull because they are sometimes null.

wsx-ucb:
Go it, thanks! I'll convert this draft once everything is finished and ask for reviews.

imkiva:
We add annotations to parameters and result types for methods that can be invoked module-wide or implemented by subclasses (in Java they are basically non-`private` ones) because it may take lots of time for people to figure out the nullability of the argument provided and avoid NPE.

Private methods usually have a smaller call scope (in Java they are callable only in themself) and most of them are just extracted utility methods, which means I can count their usage with only one hand. But it is still recommended to add annotations because IDE will warn you if you violated the nullability assumption.

base/src/main/java/org/aya/core/term/Tm.java
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * A `Tm<T>` is a more general representation of the `Term` AST.
+ * It can be better understood as representing a `Term`-like node,
+ * but with all immediate sub-`Term`s taking values from type `T`.
+ * Here we also provide casting between `Tm<Term>` and `Term` as they are isomorphic.
+ *
+ * @author wsx
+ */
+public sealed interface Tm<T> {
```
ice1000, line 26: So `Term` is just `<T extends Tm<T>> Tm<T>`?

wsx-ucb: I think you are right. In more standard terminology `Tm<T>` is the base functor of the inductive type `Term`. And you can get `Term` by getting the fixpoint of `Tm`. But I'm not sure if we need to do this in Java, so currently I just leave the definition of `Term` and `Tm` separate and only link them through the `cast` function.
This definitively leads to some code duplication, do you have any idea to improve?

ice1000: Yes, delete `Term` and refactor relevant operations. Do you have enough spare time to do it?

wsx-ucb: Yes, I can look into refactoring `Term` as well. But I'm not sure how to implement the fixpoint construction in Java?

ice1000: There is a simpler way: make `class Tm<T>` into `class Tm<T extends Tm<T>>`, and after that `Tm<?>` will be equivalent to `Term`.

wsx-ucb: Please bear with me as those Java magic is confusing me. If we define `Tm` as `sealed interface Tm<T extends Tm<T>>`, are we still allowing things like `Tm<String>` even though `String` does not extend `Tm<String>`?

ice1000: No...

wsx-ucb:
`Term.Visitor` is finally gone!
I haven't managed to do automatic code generation for `Tm<T>` and related constructs yet, but we've already got a net decrease in LOC.
Hopefully this means cleaner code for now and beyond.

ice1000:
> `Term.Visitor` is finally gone!
> I haven't managed to do automatic code generation for `Tm<T>` and related constructs yet, but we've already got a net decrease in LOC.
> Hopefully this means cleaner code for now and beyond.

Merge?

base/src/main/java/org/aya/core/term/Term.java
```diff
@@ -84,64 +174,34 @@ default VarConsumer.ScopeChecker scopeCheck(@NotNull ImmutableSeq<LocalVar> allo
    *              Can be null only if we're absolutely sure that holes are frozen,
    *              like in the error messages.
    */
-  default @NotNull Term normalize(@NotNull TyckState state, @NotNull NormalizeMode mode) {
-    if (mode == NormalizeMode.NULL) return this;
-    if (mode == NormalizeMode.NF) return this.view().normalize(state).commit();
-    return accept(new Normalizer(state), mode);
+  default @NotNull Term normalize(@Nullable TyckState state, @NotNull NormalizeMode mode) {
```
ice1000, line 177: Why nullable state?

wsx-ucb: The `normalize` method are used in places where `state` can be nullable, and the actual implementation do handle the case for null states.
I think I mislabeled it to be `NotNull` previously.

wsx-ucb:
> Merge?

I'll update the documentation a bit and mark this as ready to merge.
Do you have any suggestions on the naming of the APIs involved?
I kind of named them randomly.

wsx-ucb:
> Please make use of nullability annotations in those visitors

Now all of the `Tm` code is gone, and I've tried to add in the inferred annotation as suggested by IDEA.
If there are anything missing, please let me know or feel free to add them yourself.

base/src/main/java/org/aya/core/visitor/Expander.java
```diff
@@ -0,0 +1,171 @@
...
+          unfolded.add(fn.ref());
+          yield Expander.super.apply(fn);
+        }
+        case CallTerm.Con con -> {
+          if (!unfolding.contains(con.ref())) yield con;
+          unfolded.add(con.ref());
+          yield Expander.super.apply(con);
+        }
+        case CallTerm.Prim prim -> {
+          // TODO[kiva]: Q: is OK to use `state`? so we don't need this override.
```
ice1000, line 164: @imkiva

base/src/main/java/org/aya/core/visitor/MonoidalVarFolder.java
```diff
@@ -0,0 +1,121 @@
...
+import org.aya.core.term.*;
+import org.aya.generic.Arg;
+import org.aya.ref.DefVar;
+import org.aya.ref.Var;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.function.Function;
+
+public interface MonoidalVarFolder<R> extends Function<Term, R> {
+  // TODO: Do we need to visit variables in `access` and `new` as well?
```
ice1000, line 18: I think so.

base/src/main/java/org/aya/tyck/env/LocalCtx.java
```diff
@@ -42,16 +42,18 @@ default <T> T with(@NotNull Term.Param param, @NotNull Supplier<T> action) {
   }
   void remove(@NotNull SeqView<LocalVar> vars);
   default void forward(@NotNull LocalCtx dest, @NotNull Term term, @NotNull TyckState state) {
-    term.accept((VarConsumer<Unit>) (usage, o) -> {
-      switch (usage) {
-        case LocalVar localVar -> dest.put(localVar, get(localVar));
-        case Meta meta -> {
-          var sol = state.metas().getOrNull(meta);
-          if (sol != null) forward(dest, sol, state);
+    new VarConsumer() {
```
ice1000, line 45: I'm going to slightly change this

ice1000 (commented):
Mostly lgtm

base/src/main/java/org/aya/core/visitor/MonoidalVarFolder.java
wsx-ucb: This comment is actually from the old version of the code.
Do I need to change in this PR anyway?

base/src/main/java/org/aya/tyck/env/LocalCtx.java
wsx-ucb: Please do.

base/src/main/java/org/aya/core/visitor/Expander.java
```diff
@@ -0,0 +1,171 @@
...
+
+  record Normalizer(TyckState state) implements Expander {
+    @Override public Term post(Term term) {
+      return switch (term) {
+        case ElimTerm.App app && app.of() instanceof IntroTerm.Lambda lam -> apply(CallTerm.make(lam, app.arg()));
+        case ElimTerm.App app -> CallTerm.make(app.of(), app.arg());
+        case ElimTerm.Proj proj && proj.of() instanceof IntroTerm.Tuple tup -> {
+          var ix = proj.ix();
+          assert tup.items().sizeGreaterThanOrEquals(ix) && ix > 0 : proj.toDoc(DistillerOptions.debug()).debugRender();
+          yield apply(tup.items().get(ix - 1));
```
imkiva, line 112: ditto

image

base/src/main/java/org/aya/core/visitor/Expander.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        default -> Expander.super.post(term);
+      };
+    }
+  }
+
+  record WHNFer(TyckState state) implements Expander {
+    @Override public Term post(Term term) {
```
imkiva, line 120: annotations

base/src/main/java/org/aya/core/visitor/Expander.java
```diff
@@ -0,0 +1,171 @@
...
+          var ix = proj.ix();
+          assert tup.items().sizeGreaterThanOrEquals(ix) && ix > 0 : proj.toDoc(DistillerOptions.debug()).debugRender();
+          yield apply(tup.items().get(ix - 1));
+        }
+        default -> Expander.super.post(term);
+      };
+    }
+
+    @Override public Term apply(Term term) {
+      return switch (term) {
```
imkiva, line 134: annotations

base/src/main/java/org/aya/core/visitor/Expander.java
imkiva: Changing the `state` to `@NotNull` resolves this TODO.

base/src/main/java/org/aya/core/visitor/MonoidalVarFolder.java
```diff
@@ -0,0 +1,121 @@
...
+  R var(Var var);
+
+  R merge(SeqView<R> rs);
+
+  private SeqView<R> trace(Term term) {
```
imkiva, line 23: annotations

base/src/main/java/org/aya/core/visitor/MonoidalVarFolder.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+  private SeqView<R> trace(@NotNull Arg<Term> arg) {
+    return trace(arg.term());
+  }
+
+  @Override default R apply(Term term) {
+    return merge(trace(term));
+  }
+
+  record Usages(Var var) implements MonoidalVarFolder<Integer> {
```
imkiva, line 62: annotations

base/src/main/java/org/aya/core/visitor/TermConsumer.java
```diff
@@ -1,143 +1,30 @@
...
+ *
+ * @author wsx
+ */
+public interface TermConsumer extends Consumer<Term> {
+  default void pre(Term term) {}
+
+  default void post(Term term) {}
+
+  default void accept(Term term) {
+    pre(term);
```
imkiva, line 23: annotations

base/src/main/java/org/aya/core/visitor/VarConsumer.java
```diff
@@ -13,74 +12,25 @@
...
-    }
-
-    @Contract(pure = true) public int usageCount() {
-      return usageCount;
-    }
-
-    @Contract(mutates = "this") @Override public void visitVar(Var usage, Unit unit) {
-      if (var == usage) usageCount++;
+public interface VarConsumer extends TermConsumer {
+  default void var(Var var) {}
```
imkiva, line 16: annotation

base/src/main/java/org/aya/core/visitor/Zonker.java
```diff
@@ -55,13 +53,11 @@ public record Zonker(
     };
   }
 
-  @Override public @NotNull Term post(@NotNull Term term) {
+  @Override public Term apply(Term term) {
```
imkiva, line 56: annotations

base/src/main/java/org/aya/generic/Arg.java
```diff
@@ -19,4 +21,10 @@
   @Override public @NotNull Doc toDoc(@NotNull DistillerOptions options) {
     return Doc.bracedUnless(term.toDoc(options), explicit);
   }
+
+  public Arg<T> descent(Function<T, T> f) {
+    var term = f.apply(term());
```
imkiva, line 26: annotations

base/src/main/java/org/aya/terck/CallResolver.java
```diff
@@ -102,39 +102,16 @@ private void fillMatrix(@NotNull CallTerm callTerm, @NotNull Def callee, CallMat
...
-
-  @Override public Unit visitAccess(CallTerm.@NotNull Access term, CallGraph<Def, Term.Param> defCallGraph) {
-    resolveCall(term, defCallGraph);
-    return DefConsumer.super.visitAccess(term, defCallGraph);
-  }
-
-  @Override public Unit visitPrimCall(@NotNull CallTerm.Prim prim, CallGraph<Def, Term.Param> graph) {
-    resolveCall(prim, graph);
-    return DefConsumer.super.visitPrimCall(prim, graph);
+  @Override public void pre(Term term) {
```
imkiva, line 111: annotation

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -178,8 +178,8 @@ private void tracing(@NotNull Consumer<Trace.@NotNull Builder> consumer) {
               return fail(proj, new FieldProblem.UnknownField(proj, fieldName));
             var fieldRef = field.ref();
 
-            var structSubst = Unfolder.buildSubst(structCore.telescope(), structCall.args());
-            var tele = Term.Param.subst(fieldRef.core.selfTele, structSubst, 0);
+          var structSubst = Expander.buildSubst(structCore.telescope(), structCall.args());
+          var tele = Term.Param.subst(fieldRef.core.selfTele, structSubst, 0);
```
imkiva, line 182: indent

imkiva (changes requested):
1. Lots of annotations are missing, especially methods in interfaces. Please make use of them on-the-fly.
2. `TyckState` should always be `@NotNull` since #371.
3. There's possible duplicate but unused code in `EndoFunctor`. Please delete the unused version.
4. We have `.editorconfig` configured in the project root directory. You can try its [IDE plugin](https://editorconfig.org/#download) to avoid auto-indent using unexpected spaces.

base/src/main/java/org/aya/core/visitor/Expander.java
ice1000: But @wsx-ucb said some API passes null to it

ice1000: Can you check them for me? @imkiva

base/src/main/java/org/aya/core/visitor/MonoidalVarFolder.java
ice1000: Well let's leave it then. I think I'll handle it (or you can, if you are fast enough ;p)

base/src/main/java/org/aya/core/def/CtorDef.java
```diff
@@ -52,11 +52,7 @@ public CtorDef(
...
   }
 
-  @Override public <P, R> R accept(@NotNull Visitor<P, R> visitor, P p) {
-    return visitor.visitCtor(this, p);
-  }
-
-  public @NotNull DefVar<CtorDef, TeleDecl.DataCtor> ref() {
+    public @NotNull DefVar<CtorDef, TeleDecl.DataCtor> ref() {
     return ref;
   }
```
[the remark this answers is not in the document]
ice1000: I think it's because of some stupid code generation bugs. IDEA should auto-format the generated code, but sometimes it doesn't. I guess @wsx-ucb just needs to configure a pre-commit formatting checkbox ;p

ice1000:
@imkiva thanks!

ice1000:
Apparently, I missed every single code style problem because I was reviewing on a mobile client and the typesetting is already horrible over there 🤣

base/src/main/java/org/aya/core/visitor/Expander.java
imkiva: image

imkiva: Even in this picture, there's a `var state = new TyckState(res._1);` which should be passed to normalize

ice1000: @wsx-ucb can you fix it for us? 😉

imkiva: I guess @wsx-ucb was confused by `PrimDef.Factory.unfold` which takes a `@Nullable TyckState` and then passed it to the normalizer. But the `unfold` was always called by the normalizer with a not-null tyck state. We should make them `@NotNull` altogether.


image

ice1000: I think I'm gonna do ti

imkiva: > I think I'm gonna do ti

btw there's a duplicate but unused normalizer in `EndoFunctor`, you can also delete it

ice1000:
Reformatted!!

wsx-ucb:
@imkiva Thank you so much for the throughout code review!

> 1. Lots of annotations are missing, especially methods in interfaces and record constructors. Please make use of them on-the-fly because `@Nullable T` and `@NotNull T` are different types.

I've now added them to all the newly introduced interfaces and all implementations.

> 2. `TyckState` should always be `@NotNull` since #371.

Thanks for pointing this out.
I was a bit confused initially since all the usages of `normalize` seem to pass in a nullable state.
But now all the call sites are updated accordingly by @ice1000 in da63ea087afbd763bdcbbb4388eab0efba6fc88f.
And the implementation of normalization no longer handle null states anymore in the latest commit.

> 3. There's possible duplicate but unused code in `EndoFunctor`. Please delete the unused version.

Oops, nice catch!
Now `EndoFunctor.Normalizer` is removed.
Should be using `Expander.Normalizer`.

> 4. We have `.editorconfig` configured in the project root directory. You can try its [IDE plugin](https://editorconfig.org/#download) to avoid auto-indent using unexpected spaces.

I'm not sure when I messed up the formatting, but I'll remember to reformat my changes in the future.

ice1000:
bors d=imkiva

ice1000:
No, I can't stand it

ice1000:
bors r=[ice1000, imkiva]

ice1000 (approved):

imkiva (approved):
Thank you for the contribution of a beautiful abstraction over folding terms and being friendly to my nitpicking reviews!

Sorry for my late approval!

bors r=[ice1000, imkiva]

imkiva:
bors is not working?

bors r=[ice1000, imkiva]

imkiva:
bors cancel

imkiva:
bors r=[ice1000, imkiva]

ice1000:
@imkiva Bors *is* working. It's github who blocks the merge

PR #423, opened 2022-06-06
Classable Cherry-pick

tsao-chi:
cherry-pick from #419

PR #428, opened 2022-06-26
Flatten concrete definitions

imkiva:
in this PR, we simplified the inheritance between top-level declarations and sub-level ones introduced in #423. The approach is similar to #326 :

**TL; DR: read the code below**

- `Decl` and `Signatured` are refactored into three traits: `Telescopic`, `TopLevel` and `Resulted`
- Sub-level decls (Ctor and Field) now directly extend `Decl` (instead of `Signatured` before) like all top-level decls, the only difference is that sub-level decls do not implement the `TopLevel` trait.
- For pattern matching, `visitDecl` can now handle constructors and fields in the same `switch` clause that matches the data and struct. We don't need extra `visitCtor` and `visitField` methods anymore.
- For visiting a certain trait of one concrete definition, we can:
  - `<D extends Decl & Decl.Telescopic>` for generic functions.
  - `visitTelescopic(Decl decl, Decl.Telescopic proof)` for visitors.
- New decls should be declared in the way of:
  `[class|interface] <the-new-decl> extends CommonDecl implements <some-traits>`

See also: https://github.com/aya-prover/aya-dev/blob/4c9b212390419f234d83c51ede0e9502f7fa8e84/base/src/main/java/org/aya/concrete/stmt/Decl.java#L18-L48


ice1000 (approved):
Very good


/cc @tsao-chi

tsao-chi:
@imkiva lgtm

ice1000:
merge?

base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
```diff
@@ -24,24 +24,23 @@
...
  *
  * @author re-xyr
+ * @see Decl
  */
-public sealed abstract class TopTeleDecl extends BaseDecl.Telescopic implements TopLevelDecl {
-  public final @NotNull Accessibility accessibility;
+public sealed abstract class TeleDecl extends CommonDecl implements Decl.Telescopic, Decl.TopLevel, Decl.Resulted {
+  public final @NotNull Decl.Personality personality;
   public @Nullable Context ctx = null;
   public @NotNull Expr result;
```
ice1000, line 35: Maybe make sense to make them private

base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
```diff
@@ -24,24 +24,23 @@
...
-  public final @NotNull TopLevelDecl.Personality personality;
+  // will change after resolve
+  public @NotNull ImmutableSeq<Expr.Param> telescope;
+  public @Nullable Def.Signature signature;
```
ice1000, line 38: Ditto

base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
imkiva: Ok

base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
imkiva: These are used directly after pattern matching so it might be unnecessary to make them private. We can move to getters/setters in the future when a hook is needed.

imkiva:
bors merge

base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
ice1000: IC, thx!

PR #430, opened 2022-07-12
Fix nightly build

imkiva:
as title


imkiva:
bors merge

ice1000 (approved):
Is this graalvm action a newly appeared thing or an old thing that we didn't discover?

imkiva:
> Is this graalvm action a newly appeared thing or an old thing that we didn't discover?

Sadly it's an old thing

PR #431, opened 2022-07-12
Slightly refactor for intellij-aya

imkiva:
bors merge

### DIFF
--- a/lsp/src/main/java/org/aya/lsp/utils/ForIDEA.java
+++ b/lsp/src/main/java/org/aya/lsp/utils/ForIDEA.java
@@ -1,0 +1,19 @@
+// Copyright (c) 2020-2022 Yinsen (Tesla) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.lsp.utils;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.Serializable;
+
+/**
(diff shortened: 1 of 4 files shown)


lsp/src/main/java/org/aya/lsp/utils/ForIDEA.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.Serializable;
+
+/**
+ * Special gift for Intellij IDEA plugin. Usually used in response models to skip json serialization of certain data.
+ * @see org.aya.lsp.models.HighlightResult.Symbol
+ */
+public final class ForIDEA<T> implements Serializable {
```
ice1000, line 13: Isn't this `Ref<T>`?

ice1000 (commented):
To Kiva's question

lsp/src/main/java/org/aya/lsp/utils/ForIDEA.java
imkiva: The inner instance of `Ref<T>` is not transient. And the purpose of `ForIDEA` is skipping serialization when serving VSCode

ice1000: I would suggest asking kala-common to implement a transient Ref, what do you think

imkiva: I agree with you. /cc @Glavo

PR #432, opened 2022-07-13
Turn LSP into a pure analyzer

imkiva:
This PR
- changed LSP from a library compiler wrapper to a pure analyzer by storing compiler outputs in memory. this also boosted the LSP performance for intellij-aya.
- added a new check-only mode option (`--check`) for libraries


cli/src/main/java/org/aya/cli/utils/MainArgs.java
```diff
@@ -31,6 +31,8 @@ public static class CompileAction {
     public boolean isLibrary;
     @Option(names = {"--remake"}, description = "Treat input file as a library root and remake all")
     public boolean isRemake;
+    @Option(names = {"--check"}, description = "Treat input file as a library root and check all (no outputs will be saved to disk)")
```
ice1000, line 34: Maybe also called `--dry-run`?

ice1000: In Haskell I think there's a similar `--no-code`

cli/src/main/java/org/aya/cli/library/incremental/InMemoryCompilerAdvisor.java
```diff
@@ -0,0 +1,79 @@
...
+        .compareTo(coreLastModified.get()) > 0;
+    } catch (IOException ignore) {
+      return true;
+    }
+  }
+
+  @Override public void updateLastModified(@NotNull LibrarySource source) {
+    try {
+      coreTimestamp.put(source.file(), Files.getLastModifiedTime(source.file()));
+    } catch (IOException ignore) {
```
ice1000, line 45: Are you sure you wanna ignore that?

cli/src/main/java/org/aya/cli/library/incremental/InMemoryCompilerAdvisor.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  @Override public void prepareLibraryOutput(@NotNull LibraryOwner owner) {
+  }
+
+  @Override public void prepareModuleOutput(@NotNull LibrarySource source) {
+    // TODO: what if module name clashes?
```
ice1000, line 53: Return `bool`?

cli/src/main/java/org/aya/cli/library/incremental/InMemoryCompilerAdvisor.java
imkiva: It's safe to ignore. If the `getLastModifiedTime` failed, the `coreTimestamp` is not updated (the core file still has a smaller modified time than source file) so the corresponding source will be re-compiled next time even if it is really up-to-date.

cli/src/main/java/org/aya/cli/library/incremental/InMemoryCompilerAdvisor.java
imkiva: Module name clashing should be addressed in shallow resolver. If addressed we don't need to change anything here. I guess it's ok to return `void`

cli/src/main/java/org/aya/cli/library/incremental/InMemoryCompilerAdvisor.java
tsao-chi: How about only putting getLastModifiedTime in the try block?

imkiva: > How about only putting getLastModifiedTime in the try block?

In this case if `getLastModifiedTime` failed, we still have nothing to put. So I guess this does not matter.

cli/src/main/java/org/aya/cli/utils/MainArgs.java
imkiva: `--dry-run` means show files need processing, but do not process them. But we do tyck them.
`--no-code` looks funny and I'd like to accept it because I know [this repo](https://github.com/kelseyhightower/nocode) 🤣

imkiva:
The analysis latency was reduced by 90%!

before: 
image

after:
image

ice1000:
https://github.com/Glavo/kala-common/issues/56 /cc @imkiva

lsp/src/test/java/org/aya/lsp/LspTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Test public void testIncremental() {
+    launch(TEST_LIB).execute(
+      compile((a, e) -> {}),
+      mutate("StringPrims"),
+      compile((a, e) -> assertRemake(a, e, "StringPrims", "HelloWorld"))
+    );
+  }
+
+  @Test public void testRealWorldLike() {
```
ice1000, line 40: This is soooo dio I love that

imkiva: Thank you very much 😆

imkiva:
bors merge

PR #434, opened 2022-07-26
Upgrade kala to 0.44.0 and make some LSP APIs public

imkiva:
Also replaced deprecated usages with:
  - `kala.value.Ref` -> `kala.value.MutableValue`
  - `Option.of` -> `Option.ofNullable`


imkiva:
bors merge

PR #439, opened 2022-07-30
Fix primitive reload and polish some error messages

imkiva:
This PR also improved LSP reload performance by reducing expensive graph operations and needless LSP protocol calls.


imkiva:
bors merge

PR #441, opened 2022-08-09
Redo lift

ice1000:
Removed `RefTerm::lift`. I'm not yet sure if my changes are complete, cus in concrete there's gonna be invalid 'lift local var' and we want to report errors (now lift is restricted to defcalls).


ice1000:
I seems that the changes _are_ complete.
bors r+

PR #446, opened 2022-08-22
Partial elements

ice1000:
This PR adds partial elements to aya, renames `squeezeL` primitive to `intervalMin` and added a new primitive `intervalMax`.

Leftovers: replace cofibration syntax with interval primitives.


base/src/main/java/org/aya/concrete/Expr.java
```diff
@@ -51,7 +54,8 @@ default Expr resolve(@NotNull ModuleContext context) {
     return new ConcreteDistiller(options).term(BaseDistiller.Outer.Free, this);
   }
 
-  @ForLSP sealed interface WithTerm extends Expr {
+  @ForLSP
+  sealed interface WithTerm extends Expr {
```
ice1000, line 58: :trollface:

base/src/main/java/org/aya/concrete/Expr.java
```diff
@@ -251,6 +255,34 @@ record BinOpSeq(
     @NotNull ImmutableSeq<NamedArg> seq
   ) implements Expr {}
 
+  record FaceExpr(@NotNull SourcePos sourcePos) implements Expr {}
+
+  /** partial element */
```
ice1000, line 260: Needs update, see https://github.com/ice1000/guest0x0/compare/6b0196b6b21ec6a9086edb97802f21bbafbd10f7...415744a9ed74e825bd88185debfd6999d131c35c

ice1000 (commented):
incomplete

base/src/main/java/org/aya/core/def/PrimDef.java
```diff
@@ -246,8 +205,6 @@ public void clear(@NotNull ID name) {
   public enum ID {
     /** Short for <em>Arend coe</em>. */
     ARCOE("arcoe"),
-    SQUEEZE_LEFT("squeezeL"),
```
ice1000, line 249: I think these functions can still be primitives, so they can be renamed. We can transform them into special ASTs in elaboration and leave the reduction function here as the identity function

base/src/main/java/org/aya/concrete/Expr.java
imkiva: woww!! with this change I finally understand the code below.
image

imkiva: No I don't understand

ice1000: Bc type checking guarantees both sides to be partial elements of the same type (and have the same set of faces). It remains to make sure all faces are the same. So, we just traverse the clauses of one side and see if the other side reduces the same way.

imkiva: Thank you for the explanation. But my problem was the `n` here is also a partial element (it should be the right-hand side of another partial element). This eventually compares `u` with `{| u |}`. Am I wrong? 😭

ice1000: No, you're right. I'll fix it in guest. Thanks for the reminder!

base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -137,6 +138,12 @@ yield visitCalls(false, fn, (nc, l) -> l.toDoc(options), outer,
         .map($ -> Doc.styled(KEYWORD, Doc.symbol("ulift")))
         .appended(term(Outer.Lifted, expr.expr())));
       case Expr.MetaPat metaPat -> metaPat.meta().toDoc(options);
+      case Expr.PartTy ty -> Doc.sep(Doc.styled(KEYWORD, "Partial"),
+        ty.type().toDoc(options), Doc.braced(ty.restr().toDoc()));
+      case Expr.PartEl el -> Doc.sep(Doc.symbol("{|"),
+        Doc.join(Doc.symbol(" | "), el.clauses().map(Restr.Side::toDoc)),
```
ice1000, line 144: `Doc.spaced(Doc.symbol("|"))` maybe

base/src/main/java/org/aya/distill/CoreDistiller.java
```diff
@@ -159,6 +159,13 @@ yield visitCalls(false, term(Outer.AppHead, head), args.view(), outer,
           : linkLit(shaped.repr(), suc.ref, CON_CALL),
         () -> Doc.plain(String.valueOf(shaped.repr())));
       case PrimTerm.Str str -> Doc.plain("\"" + StringEscapeUtil.escapeStringCharacters(str.string()) + "\"");
+      case FormTerm.PartTy ty -> Doc.sep(Doc.styled(KEYWORD, "Partial"),
+        term(Outer.AppSpine, ty.type()), Doc.braced(ty.restr().toDoc()));
+      case IntroTerm.HappyPartEl el -> Doc.sep(Doc.symbol("{|"),
+        Doc.join(Doc.symbol(" | "), el.clauses().map(Restr.Side::toDoc)),
```
ice1000, line 165: Ditto

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -249,10 +253,74 @@ yield new Result(new IntroTerm.Tuple(items.map(Result::wellTyped)),
...
+    }
+  }
+
+  private @NotNull Option<Restr.Side<Term>> clause(@NotNull Restr.Side<Expr> clause, @NotNull Term type) {
+    var cofib = new Restr.Cofib<>(clause.cof().ands().map(this::condition));
+    var u = CofThy.vdash(cofib, new Subst(), subst ->
+      inherit(clause.u(), subst.term(state, type)).wellTyped);
+    if (u.isDefined() && u.get() == null) {
+      // ^ some `inst` in `cofib.ands()` are ErrorTerms.
+      // Q: report error again?
```
ice1000, line 318: I think we don't have to.

base/src/main/java/org/aya/core/visitor/Expander.java
```diff
@@ -105,9 +137,33 @@ record Normalizer(@NotNull TyckState state) implements Expander {
...
+              yield new IntroTerm.SadPartEl(u);
+            }
+          }
+          yield new IntroTerm.HappyPartEl(clauses.toImmutableSeq());
+        }
+      };
+    }
   }
 
   record WHNFer(@NotNull TyckState state) implements Expander {
```
wsx-ucb, line 169: Should we also change the logic for `WHNFer`?
Or even unify most of that with `Normalizer`?

base/src/main/java/org/aya/tyck/LittleTyper.java
```diff
@@ -73,10 +75,20 @@ public record LittleTyper(@NotNull TyckState state, @NotNull LocalCtx localCtx)
...
       }
       case FormTerm.Univ univ -> new FormTerm.Univ(univ.lift() + 1);
-      case FormTerm.Interval interval -> new FormTerm.Univ(0);
-      case PrimTerm.End end -> FormTerm.Interval.INSTANCE;
+      case FormTerm.Interval interval -> FormTerm.Univ.ZERO;
+      case PrimTerm.Mula end -> FormTerm.Interval.INSTANCE;
       case PrimTerm.Str str -> state.primFactory().getCall(PrimDef.ID.STR);
       case LitTerm.ShapedInt shaped -> shaped.type();
+      case FormTerm.PartTy ty -> FormTerm.Univ.ZERO;
+      case IntroTerm.HappyPartEl el -> {
```
ice1000, line 83: Hmm, should we store `A` in the partial term?

ice1000: It can happen that we are unable to infer the type `A`.

base/src/main/java/org/aya/core/visitor/Expander.java
ice1000: Partial is already a whnf, I think it's fine.

base/src/main/java/org/aya/tyck/LittleTyper.java
imkiva: > It can happen that we are unable to infer the type `A`.

I guess this only happens when a partial element doesn't have clauses? Currently we don't allow this case so maybe it's not needed yet?

base/src/main/java/org/aya/tyck/unify/DefEq.java
```diff
@@ -299,35 +299,8 @@ private boolean doCompareTyped(@NotNull Term type, @NotNull Term lhs, @NotNull T
...
-                return compare(lu, ru, lr, rl, subst.term(state, A));
-              })));
-          case IntroTerm.HappyPartEl lp && rel instanceof IntroTerm.SadPartEl rp -> comparePartial(lp, rp, lr, rl, A);
-          case IntroTerm.SadPartEl lp && rel instanceof IntroTerm.HappyPartEl rp -> comparePartial(rp, lp, rl, lr, A);
-          // ^ should swap lr and rl?
-          default -> false;
-        };
-      }
+      case FormTerm.PartTy ty -> CofThy.conv(ty.restr(), new Subst(),
+        subst -> doCompareTyped(ty.type(), lhs, rhs, lr, rl));
```
imkiva, line 303: This looks incorrect. We should compare `lhs.u()` with `rhs.u()`

imkiva: I am temporarily reverting this change bc I need to test generalized path. 😉

base/src/main/java/org/aya/tyck/LittleTyper.java
ice1000: What if we have no clause in the partial at all?

base/src/main/java/org/aya/tyck/unify/DefEq.java
ice1000: You're right about this code being wrong, but I have another fix, hold on

base/src/main/java/org/aya/tyck/LittleTyper.java
imkiva: Then its type `Partial A r` can never reduce to `A` since we don't know if `isOne r` holds.

imkiva: Oh it seems agda allows empty partials

ice1000: I actually think empty partials are harmful. But we should store the type anyway, bc the boundaries can have different types

imkiva:
bors merge

ice1000:
@imkiva Wait!!
bors r-

imkiva:
Ok, now
bors merge

imkiva:
> Build succeeded:
> 
> * [gradle-check](https://github.com/aya-prover/aya-dev/runs/8038589332?check_suite_focus=true)

Thank you very much!!

Issue #448, opened 2022-08-24
Is it useful to implement `isType` for universe-polymorphic checking of a type?


ice1000:
Currently if we know a term `a` should be a type, the only thing we can do is either:

+ `inherit(a, FormTerm.Univ.ZERO)`, or
+ `synthesize(a)` and check if the returned type is `FormTerm.Univ`

I think neither is good, but I can't say for sure. Maybe we should have an inherit-like function for types, without explicitly asking which level it should live in. This is also a preparation for introducing `Set` which is a different universe (and is polymorphic) and should be distinguished from `Type`.

ice1000:
@tsao-chi @lunalunaa feel free to take over

PR #468, opened 2022-08-31
Generic FormTerm.Univ check, able to solve a lifted meta

tsao-chi:
This might be one possible solution to https://github.com/aya-prover/aya-dev/issues/448

### DIFF
--- a/base/src/main/java/org/aya/core/term/FormTerm.java
+++ b/base/src/main/java/org/aya/core/term/FormTerm.java
@@ -58,6 +58,10 @@ public sealed interface FormTerm extends Term {
    */
   record Univ(int lift) implements FormTerm {
     public static final @NotNull FormTerm.Univ ZERO = new Univ(0);
+
+    // Only used in comparison. Invalid in other places.
+    public static final int UNKNOWN_LIFT = Integer.MIN_VALUE;
+    public static final @NotNull FormTerm.Univ UNKNOWN = new Univ(UNKNOWN_LIFT);
   }
 
   final class Interval implements FormTerm {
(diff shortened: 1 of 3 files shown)


ice1000:
I think this solution has a problem. Let's say we have a hole, and we compare it with unknown level. This gives us core term with unknown levels.



If I'm not wrong on assuming unknown level represents "infinity level", this can generate a Girard paradox

ice1000:
I think this solution has a problem. Let's say we have a hole, and we compare it with unknown level. This gives us core term with unknown levels.



If I'm not wrong on assuming unknown level represents "infinity level", this can generate a Girard paradox

ice1000:
But unknown levels are going to be in the core, am I right?

tsao-chi:
> But unknown levels are going to be in the core, am I right?

This should be prevented

ice1000:
> > But unknown levels are going to be in the core, am I right?
> 
> This should be prevented

I think we should prevent using invalid core term for the sake of type checking.

We are inevitably going to have a isType function when we introduce more universes.

tsao-chi:
> I think we should prevent using invalid core term for the sake of type checking.

I think it may be fine if it is invisible to users and other parts of aya

ice1000:
> > I think we should prevent using invalid core term for the sake of type checking.
> 
> I think it may be fine if it is invisible to users and other parts of aya

But it's visible to type checking, and I think it's dangerous (while having a separate function for checking types is easy and harmless and not so much more difficult and is going to be done anyway).

You may solve metas to a unknown level, this can be an intermediate term used in type checking that does not get into the results, and this can make the results invalid.

Also, this is going to be very difficult to reimplement Aya in Aya in the future because we need to deal with unknown levels everywhere, like in serialization or something.

Most importantly, when we introduce more universes, this hack will become completely unnecessary :)

ice1000:
I'll create a new issue for you and you'll see why a separate function is necessary

ice1000:
Also, how is this related to lifted meta?

ice1000:
> I'll create a new issue for you and you'll see why a separate function is necessary

#469

tsao-chi:
Closing this hack

ice1000:
Are you implementing #469?

tsao-chi:
@ice1000 I will implement #448 first

PR #474, opened 2022-09-06
TYCK: add inheritUniv() api

tsao-chi:
This is a part of https://github.com/aya-prover/aya-dev/issues/448

### DIFF
--- a/base/src/main/java/org/aya/tyck/ExprTycker.java
+++ b/base/src/main/java/org/aya/tyck/ExprTycker.java
@@ -504,7 +504,6 @@ public final class ExprTycker extends Tycker {
     return switch (expr) {
       case Expr.TupExpr tuple -> fail(tuple, univ, BadTypeError.sigmaCon(state, tuple, univ));
       case Expr.HoleExpr hole -> {
-        // TODO[lift-meta]: should be able to solve a lifted meta
         var freshHole = localCtx.freshHole(univ, Constants.randomName(hole), hole.sourcePos());
         if (hole.explicit()) reporter.report(new Goal(state, freshHole._1, hole.accessibleLocal().get()));
         yield new Result(freshHole._2, univ);
--- a/base/src/main/java/org/aya/tyck/ExprTycker.java
+++ b/base/src/main/java/org/aya/tyck/ExprTycker.java
@@ -75,13 +75,11 @@ public final class ExprTycker extends Tycker {
         var param = pi.param();
         final var var = param.ref();
         var domTy = param.type();
-        var domRes = synthesize(domTy);
-        var domLvl = ensureUniv(domTy, domRes.type());
+        var domRes = universe(domTy);
         var resultParam = new Term.Param(var, domRes.wellTyped(), param.explicit());
         yield localCtx.with(resultParam, () -> {
-          var cod = synthesize(pi.last());
-          var codLvl = ensureUniv(pi.last(), cod.type());
-          return new TermResult(new FormTerm.Pi(resultParam, cod.wellTyped()), new FormTerm.Univ(Math.max(domLvl, codLvl)));
+          var cod = universe(pi.last());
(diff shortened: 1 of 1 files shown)


ice1000 (commented):
+ `inheritUniv` -> `universe` (and `doUniverse`), and if you hate this exaggerated name you can rename it everywhere to `Sort` I guess (just make sure the naming is consistent, I don't mind huge rename commits, but please submit standalone PRs for renames)
+ It doesn't have to return a `Result`. I think we only need a `Term`. I think the same applies to `inherit`, but I'm unsure why we're keeping it. Maybe it's time to change.
+ Rest lgtm

ice1000 (commented):
+ inheritUniv -> universe (and doUniverse), and if you hate this exaggerated name you can rename it everywhere to Sort I guess (just make sure the naming is consistent, I don't mind huge rename commits, but please submit standalone PRs for renames)
+ It doesn't have to return a Result. I think we only need a Term. I think the same applies to inherit, but I'm unsure why we're keeping it. Maybe it's time to change.
+ universe should take an optional upper bound. Since this is a parameter and is not passed to the constructor of FormTerm.Univ it is okay to use -1 this time.
+ Rest lgtm

tsao-chi:
> It doesn't have to return a Result. I think we only need a Term. I think the same applies to inherit, but I'm unsure why we're keeping it. Maybe it's time to change.

I think Result might be necessary because these 2 functions do process type information

ice1000:
@tsao-chi You're right.

What about using -1? I think it should be more memory efficient.

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -497,6 +497,33 @@ yield switch (term.normalize(state, NormalizeMode.WHNF)) {
     };
   }
 
+  private @NotNull Result doUniverse(@NotNull Expr expr, @NotNull Option<Integer> upperBound) {
+    if (upperBound.isDefined()) assert (upperBound.get() >= 0);
+    Term univ = FormTerm.Univ.ZERO;
+    return switch (expr) {
+      case Expr.TupExpr tuple -> fail(tuple, univ, BadTypeError.sigmaCon(state, tuple, univ));
+      case Expr.HoleExpr hole -> {
+        // TODO[lift-meta]: should be able to solve a lifted meta
```
ice1000, line 506: We no longer need it

ice1000 (commented):
Can we use `universe` in synth and inherit?

tsao-chi:
@tsao-chi

ice1000 (commented):
Did you implement `doUniverse(Pi)` and `doUniverse(Sigma)` and `doUniverse(DataCall)? Those are needed

tsao-chi:
> Did you implement `doUniverse(Pi)` and `doUniverse(Sigma)` and `doUniverse(DataCall)? Those are needed

No, synthesize() are re-used for those terms

ice1000 (commented):
I'm talking about a function `f(u)` that checks if `u` is a valid type, so apart from universes, you also move the logic of checking `Pi`/`Sigma` into the new function as well. Also the return type semantics needs to be changed: it's the level of the type of the type, not the type itself.

ice1000:
> > Did you implement `doUniverse(Pi)` and `doUniverse(Sigma)` and `doUniverse(DataCall)? Those are needed
> 
> No, synthesize() are re-used for those terms

Can you put everything together? Because synthesize don't work for `Pi`. In fact, `Pi` is overloaded:

+ `A -> B : Type` for `A, B : Type` or `A : ISet`
+ `A -> B : ISet` for `A, B : ISet`
+ `A -> B : Set` for `A, B : Set`

We need `universe` to handle all these

ice1000:
Note that `Set` and `Type` are both universe polymorphic

ice1000:
Gooooooooooooooooooooooood

PR #475, opened 2022-09-06
REFACTOR: extract resolveOpDecl()

tsao-chi:


base/src/main/java/org/aya/ref/DefVar.java
```diff
@@ -61,6 +61,10 @@ private DefVar(Concrete concrete, Core core, @NotNull String name) {
     return new DefVar<>(null, null, name);
   }
 
+  public static OpDecl resolveOpDecl(@NotNull DefVar<?,?> defVar, @NotNull ImmutableSeq<String> moduleName){
+    return defVar.opDeclRename.getOrDefault(moduleName, defVar.opDecl);
+  }
+
```
imkiva, line 67: It is not needed to be static. And it would be nice if `@Nullable OpDecl`

imkiva (approved):
bors d=tsao-chi

imkiva (commented):
Should we also change `DefVar.isInfix()`? It is still not checking `opDeclRename`

base/src/main/java/org/aya/ref/DefVar.java
tsao-chi: resolved

tsao-chi:
> ~Should we also change `DefVar.isInfix()`? It is still not checking `opDeclRename`~
> 


We don't have module information in `isInfix` usages

base/src/main/java/org/aya/ref/DefVar.java
```diff
@@ -61,6 +61,10 @@ private DefVar(Concrete concrete, Core core, @NotNull String name) {
     return new DefVar<>(null, null, name);
   }
 
+  public @Nullable OpDecl resolveOpDecl(@NotNull ImmutableSeq<String> moduleName){
```
ice1000, line 64: `@NotNull`?

ice1000: @imkiva are you sure it's `@Nullable`?

ice1000: Ok I was stupid

ice1000:
bors r=tsao-chi

Issue #472, opened 2022-09-01
Pretty printer bug


imkiva:
This assertion
```patch
Index: base/src/test/java/org/aya/core/DistillerTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/base/src/test/java/org/aya/core/DistillerTest.java b/base/src/test/java/org/aya/core/DistillerTest.java
--- a/base/src/test/java/org/aya/core/DistillerTest.java	(revision 3275eeb0b50bc6cec23bbdacf86aa6c734cc02cb)
+++ b/base/src/test/java/org/aya/core/DistillerTest.java	(date 1662024364066)
@@ -88,6 +88,18 @@
     assertEquals("(Pi (A : Type 0) -> A) = X", test2.toDoc(DistillerOptions.informative()).debugRender());
   }
 
+  @Test public void binop() {
+    var decls = TyckDeclTest.successTyckDecls("""
+      open data Nat | zero | suc Nat
+      open data D   | infix · Nat Nat
+      
+      def g (h : Nat -> D) : Nat => zero
+      def t (n : Nat) => g (n ·)
+      """)._2;
+    var t = ((FnDef) decls.get(3)).body.getLeftValue();
+    assertEquals("g (n ·)", t.toDoc(DistillerOptions.informative()).debugRender());
+  }
+
   private @NotNull Doc declDoc(@Language("TEXT") String text) {
     return Doc.vcat(TyckDeclTest.successTyckDecls(text)._2.map(d -> d.toDoc(DistillerOptions.debug())));
   }

```

would fail:

```
org.opentest4j.AssertionFailedError: 
Expected :g (n ·)
Actual   :g · n
```

PR #476, opened 2022-09-06
PRETTY: fix partially applied binop

tsao-chi:
fix https://github.com/aya-prover/aya-dev/issues/472


imkiva (approved):
bors d=tsao-chi

ice1000 (approved):
bors r+

PR #478, opened 2022-09-07
Increase test coverage, remove unused classes

ice1000:
Just trying


base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -511,11 +510,20 @@ yield switch (term.normalize(state, NormalizeMode.WHNF)) {
...
+          case CallTerm.Hole hole -> {
+            unifyTyReported(hole, FormTerm.Univ.ZERO, expr);
+            yield 0;
+          }
+          default -> {
+            reporter.report(BadTypeError.univ(state, expr, lower));
+            yield 0;
+          }
+        };
+        yield new UnivResult(result.wellTyped(), lvl);
```
imkiva, line 526: duplicates with StmtTycker?

ice1000: Thank you

PR #481, opened 2022-09-08
Be strict when comparing constructors for termination

### LINKED ISSUE #471: Termination bug
Related: https://github.com/agda/agda/issues/6059

```aya
-- adapted from: https://github.com/agda/agda/blob/master/test/Fail/Issue6059.agda

open data Nat | zero | suc Nat
open data D   | infix · Nat Nat

-- The termination checker faultily accepts this call pattern

module Bug {
  def f D : Nat
    | suc n · _ => f (n · 2)                      -- this call to f is fishy!
    | n · suc (suc m) => f (suc n · m)
    | 0 · 0 => 0
    | 0 · 1 => 0
}

-- but if you extract `f (n · 2)` to `g`, the loop is detected:

module Expected {
  def f D : Nat
    | suc n · _ => g (n ·)
    | n · suc (suc m) => f (suc n · m)
    | 0 · 0 => 0
    | 0 · 1 => 0

  def g (h : Nat -> D) : Nat => f (h 2)
}

```

imkiva:
fix #471

After days of digging into agda's codebase (most of the time was spent on setting up a haskell ide) I finally know the differences between foetus in paper and foetus in agda. I decide to bring them to aya after composing some conclusion notes.

The `FoetusLimitation.aya` test is only removed temporarily. We will see it back sooner or later 😉

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -37,8 +37,6 @@ However, we can share some cool stuffs here:
   (instead of a number, such as in Haskell or Agda)
   which is useful for [equation reasoning][assoc].
 + Termination checker inspired from foetus.
-  We adapted some code from Agda's implementation to accept
-  [more definitions][foetus] (which are rejected by, e.g. Arend).
 + Inference of type checking order. That is to say,
   no syntax for forward-declarations is needed for [mutual recursions][mutual].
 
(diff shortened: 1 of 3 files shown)


ice1000:
Ok

bors r+

PR #482, opened 2022-09-08
Consider size-change in termination checker

### LINKED ISSUE #471: Termination bug

imkiva:
This is highly based on the idea from:

- Amir M. Ben-Amram, & Chin Soon Lee (2009). Ranking Functions for Size-Change Termination II. Log. Methods Comput. Sci., 5(2).

and implementation details from:

- https://github.com/agda/agda/blob/master/src/full/Agda/Termination/TermCheck.hs

Really fixes #471 and supersedes #481

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -37,6 +37,8 @@ However, we can share some cool stuffs here:
   (instead of a number, such as in Haskell or Agda)
   which is useful for [equation reasoning][assoc].
 + Termination checker inspired from foetus.
+  We adapted some code from Agda's implementation to accept
+  [more definitions][foetus] (which are rejected by, e.g. Arend).
 + Inference of type checking order. That is to say,
   no syntax for forward-declarations is needed for [mutual recursions][mutual].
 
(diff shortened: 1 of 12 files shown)


ice1000 (commented):
What about some tests that pretty prints the call matrices? I'm curious on what they look like

imkiva:
> What about some tests that pretty prints the call matrices? I'm curious on what they look like

They keep a distance of 3-spaces from each other (1-space before)

image

ice1000 (approved):
bors r+

PR #484, opened 2022-09-10
Reimplement call graph completion, supports interleaved arguments on recursion

imkiva:
Reference:
- [CallGraph.hs#L155](https://github.com/agda/agda/blob/e3bf58d8b2e95bc0481035756f44ddd9fe19b40d/src/full/Agda/Termination/CallGraph.hs#L155)
- [TermCheck.hs#L296](https://github.com/agda/agda/blob/e3bf58d8b2e95bc0481035756f44ddd9fe19b40d/src/full/Agda/Termination/TermCheck.hs#L296)


base/src/main/java/org/aya/terck/CallGraph.java
```diff
@@ -20,72 +22,126 @@
  * @see CallMatrix
  */
 public record CallGraph<T, P>(
```
ice1000, line 24: It might be a good idea to move this to `tools`?

imkiva: Yes this is planned

imkiva: So other projects (like guest, anqur) can have termination check for free!

imkiva:
bors merge

imkiva:
image

imkiva:
bors merge

PR #485, opened 2022-09-10
Update bors message, upgrade guest

ice1000:
bors r+

/cc @lunalunaa the function you need is `CofThy.isOne`


imkiva:
Should bors respect commit check as well?

ice1000:
@imkiva I agree

lunalunaa:
I wonder why it has this name though

> bors r+
> 
> /cc @lunalunaa the function you need is `CofThy.isOne`

ice1000:
@lunalunaa > I wonder why it has this name though 

It's a reference to Cubical Agda's IsOne

PR #487, opened 2022-09-10
IntelliJ theme

ice1000:
As said in the commit messages, I've added a new theme and it's randomly used in the codebase now.

I've refactored `PrinterConfig` and inheritors so that color schemes and styles are now taken from some constructor calls instead of in the deepest constructor overload of `Stylist`.

Anybody wanna make the color themes configurable?

bors r+


ice1000 (commented):
bors r-

ice1000 (commented):
bors r+

PR #488, opened 2022-09-10
Distinguish unary from binary

### LINKED ISSUE #462: Overhual binop
Current:
- `def infix` handles all operators: unary and binary

Wanted:
- `def infix` handles all binary operators
- `def fixl`, `def fixr` for all unary operators
- Strict argument count assumption and section syntactic sugar

imkiva:
Fix #462 and removed `bind` keyword

before: `def x => {??} bind looser +` => now: `def x => {??} looser +`
before: `def x => {??} bind { looser a tighter b }` => now: `def x => {??} looser a tighter b`

### DIFF
--- a/base/src/main/java/org/aya/tyck/unify/DefEq.java
+++ b/base/src/main/java/org/aya/tyck/unify/DefEq.java
@@ -16,6 +16,7 @@ import org.aya.core.term.*;
 import org.aya.core.visitor.Expander;
 import org.aya.core.visitor.Subst;
 import org.aya.generic.Arg;
+import org.aya.generic.Cube;
 import org.aya.generic.util.InternalException;
 import org.aya.generic.util.NormalizeMode;
 import org.aya.guest0x0.cubical.CofThy;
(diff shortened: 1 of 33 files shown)


imkiva:
We need path type conversion to accept this code

image

base/src/main/java/org/aya/tyck/unify/DefEq.java
```diff
@@ -303,19 +304,47 @@ private boolean doCompareTyped(@NotNull Term type, @NotNull Term lhs, @NotNull T
...
+      default -> false;
+    };
+  }
+
+  private boolean compareCube(@NotNull Cube<Term> lhs, @NotNull Cube<Term> rhs, Sub lr, Sub rl) {
+    lhs.params().zipView(rhs.params()).forEach(x -> {
+      lr.map.put(x._1, new RefTerm(x._2));
+      rl.map.put(x._2, new RefTerm(x._1));
+    });
+    // TODO: let CofThy.propExt uses lr and rl?
```
ice1000, line 334: Can we just capture these variables in the lambda in propExt?

tools/src/main/java/org/aya/util/binop/Assoc.java
```diff
@@ -1,25 +1,42 @@
...
-  /**
-   * Parenthesis operator fixity.
-   */
-  Infix(true),
-  InfixL(true),
-  InfixR(true),
-  Invalid(false);
+  /** infix */
+  Infix,
+  /** infix, but associates leftly */
```
ice1000, line 11: left-associative

base/src/main/java/org/aya/tyck/unify/DefEq.java
imkiva: What do you mean by "capture" (sorry I didn't get it)

imkiva: We can address this later :)

imkiva:
bors merge

PR #489, opened 2022-09-11
Making `I`, and `Partial` a primitive

lunalunaa:
`I` and `Partial` are now a primitive instead of a keyword.

## What still needs to be done
- [x] Modifying test cases
- [ ] ~~Elaborate primitives at `CallTerm.make`~~

Edit (@ice1000): we no longer hack `CallTerm.make`, but we instead hack `ExprTycker::defCall`, and we eagerly reduce primCalls.

In the future, we will ban functions that takes `Type` types into `ISet` types, and we only allow pure `ISet` functions. Those functions will be marked as always-inline.

### DIFF
--- a/base/src/main/java/org/aya/core/def/PrimDef.java
+++ b/base/src/main/java/org/aya/core/def/PrimDef.java
@@ -170,8 +170,8 @@ public final class PrimDef extends TopLevelDef {
       public final @NotNull PrimDef.PrimSeed PARTIAL =
         new PrimSeed(ID.PARTIAL,
           (prim, state) -> {
-            var ty = prim.args().get(0).term().normalize(state, NormalizeMode.WHNF);
-            var iExp = prim.args().get(1).term().normalize(state, NormalizeMode.WHNF);
+            var iExp = prim.args().get(0).term().normalize(state, NormalizeMode.WHNF);
+            var ty = prim.args().get(1).term().normalize(state, NormalizeMode.WHNF);
 
             return new FormTerm.PartTy(ty, isOne(iExp));
           },
--- a/base/src/main/java/org/aya/core/def/PrimDef.java
+++ b/base/src/main/java/org/aya/core/def/PrimDef.java
@@ -167,7 +167,7 @@ public final class PrimDef extends TopLevelDef {
 
       public final @NotNull PrimDef.PrimSeed PARTIAL =
         new PrimSeed(ID.PARTIAL,
-          ((prim, state) -> {
+          (prim, state) -> {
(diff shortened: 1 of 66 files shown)


base/src/main/java/org/aya/core/def/PrimDef.java
```diff
@@ -156,6 +160,29 @@ private final class Initializer {
...
+        new PrimSeed(ID.I,
+          ((prim, tyckState) -> PrimTerm.Interval.INSTANCE),
+          ref -> new PrimDef(ref, FormTerm.Univ.ZERO, ID.I),
+          ImmutableSeq.empty());
+
+      public final @NotNull PrimDef.PrimSeed PARTIAL =
+        new PrimSeed(ID.PARTIAL,
+          (prim, state) -> {
+            var ty = prim.args().get(0).term().normalize(state, NormalizeMode.WHNF);
+            var iExp = prim.args().get(1).term().normalize(state, NormalizeMode.WHNF);
```
ice1000, line 174: These code are unreachable

lunalunaa: outdated

base/src/main/java/org/aya/core/def/PrimDef.java
```diff
@@ -156,6 +160,29 @@ private final class Initializer {
 
         throw new InternalException("Expected strings but found " + first + " and " + second);
       }
+
+      public final @NotNull PrimDef.PrimSeed I =
+        new PrimSeed(ID.I,
+          ((prim, tyckState) -> PrimTerm.Interval.INSTANCE),
```
ice1000, line 166: This function will never be called. I recommend using `throw AssertionError` here

base/src/main/java/org/aya/core/def/PrimDef.java
```diff
@@ -156,6 +160,29 @@ private final class Initializer {
...
+          ref -> new PrimDef(ref, FormTerm.Univ.ZERO, ID.I),
+          ImmutableSeq.empty());
+
+      public final @NotNull PrimDef.PrimSeed PARTIAL =
+        new PrimSeed(ID.PARTIAL,
+          (prim, state) -> {
+            var iExp = prim.args().get(0).term().normalize(state, NormalizeMode.WHNF);
```
ice1000, line 173: Ditto

base/src/main/java/org/aya/core/def/PrimDef.java
imkiva: Yes please do the same for imin and imax

imkiva (approved):
Rua

PR #490, opened 2022-09-11
Clean up

imkiva:
- Reorganize problems by their type
- Less visitors
- Delete really unused classes & methods


base/src/main/java/org/aya/resolve/error/PrimResolveError.java
```diff
@@ -0,0 +1,55 @@
...
+  @Override default @NotNull Stage stage() {return Stage.RESOLVE;}
+  @Override default @NotNull Severity level() {return Severity.ERROR;}
+
+  record UnknownPrim(
+    @Override @NotNull SourcePos sourcePos,
+    @NotNull String name
+  ) implements PrimResolveError {
+    @Override public @NotNull Doc describe(@NotNull DistillerOptions options) {
+      return Doc.sep(
+        Doc.plain("Unknown primitive"),
```
ice1000, line 24: English

ice1000 (commented):
Excellent!

Are we eventually going to use the cancellation indicators? I still kinda think they're useful. Maybe it's because we didn't have very long code at this point.

imkiva:
> Excellent!
> 
> Are we eventually going to use the cancellation indicators? I still kinda think they're useful. Maybe it's because we didn't have very long code at this point.

I don't know either... but they are more useful as UI actions rather than here (since you can always use ctrl-c for cmdline apps).

base/src/main/java/org/aya/resolve/error/PrimResolveError.java
imkiva: nice catch~

ice1000:
@imkiva Yes but what about LSP

imkiva:
> @imkiva Yes but what about LSP

It applies the way JVM terminates a completable future.

ice1000:
bors r_

imkiva:
bors cancel

imkiva:
bors merge

PR #491, opened 2022-09-12
Some changes

ice1000:
@imkiva

### DIFF
--- a/base/src/main/java/org/aya/core/term/Term.java
+++ b/base/src/main/java/org/aya/core/term/Term.java
@@ -122,15 +122,8 @@ public sealed interface Term extends AyaDocile, Restr.TermLike<Term> permits Cal
       }
       case CallTerm.Prim prim -> {
         var args = prim.args().map(arg -> arg.descent(f));
-        yield switch (prim.id()) {
-          case IMIN -> PrimTerm.Mula.and(args.first().term(), args.last().term());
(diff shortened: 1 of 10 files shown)


base/src/main/java/org/aya/core/term/Term.java
```diff
@@ -122,15 +122,8 @@ public sealed interface Term extends AyaDocile, Restr.TermLike<Term> permits Cal
...
-          case IMIN -> PrimTerm.Mula.and(args.first().term(), args.last().term());
-          case IMAX -> PrimTerm.Mula.or(args.first().term(), args.last().term());
-          case INVOL -> PrimTerm.Mula.inv(args.first().term());
-          default -> {
-            if (args.sameElements(prim.args(), true)) yield prim;
-            yield new CallTerm.Prim(prim.ref(), prim.ulift(), args);
-          }
-        };
+        if (args.sameElements(prim.args(), true)) yield prim;
+        yield new CallTerm.Prim(prim.ref(), prim.ulift(), args);
```
imkiva, line 126: What if the args are not reduced but the head is an interval call?

imkiva: No.

ice1000: Most likely. But how to deal with it?

imkiva: I don't think it's here bc unfolding happens in `Expander` after this descent

PR #526, opened 2022-09-19
Distiller fix

### LINKED ISSUE #514: Pretty print precedence for interval expressions
image

Should have parenthesis

### LINKED ISSUE #515: Pretty print for pathapp/pathlam is huacked up
`PathApp(Lam(x, a), b)` rendered as `\x. a b` what the

wsx-ucb:
Attempt to fix #514 and fix #515.

### DIFF
--- a/base/src/main/java/org/aya/distill/BaseDistiller.java
+++ b/base/src/main/java/org/aya/distill/BaseDistiller.java
@@ -124,7 +124,6 @@ public abstract class BaseDistiller<Term extends AyaDocile> {
   }
 
   public static @NotNull Doc checkParen(@NotNull Outer outer, @NotNull Doc binApp, @NotNull Outer binOp) {
-    if (outer == binOp && (outer == Outer.IMin || outer == Outer.IMax)) return binApp;
     return outer.ordinal() >= binOp.ordinal() ? Doc.parened(binApp) : binApp;
   }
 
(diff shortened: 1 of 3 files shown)


wsx-ucb:
Fixing the issue requires passing around the appropriate `Outer` parameter.
I find it more convenient to achieve it using non-static method and calling `term` directly.
Are there any reason why helper methods like `formula`, `partial`, `cube` are declared as static previously?

ice1000:
@wsx-ucb I think it's fine to instance method them. Have you tried f (a and b and c)? Will there be a paren

ice1000:
@wsx-ucb > I'll then try to do the same thing for partial, cube, etc.? 

I think it unnecessary.

And yes I think we should. You may want to add some enum variants to Prec and don't worry about that. It's very flexible interface

ice1000:
@ice1000 Sorry it should be Outer not Prec

imkiva:
> Are there any reason why helper methods like `formula`, `partial`, `cube` are declared as static previously?

No specific reasons, but I guess that's just because we want them to compile at that time

wsx-ucb:
> You may want to add some enum variants to Outer and don't worry about that.

Is it really this simple?
Currently `Outer` is used in `checkParen` such that only when we have strictly looser bindings the parenthesis would be omitted.
Also, the `Outer` is setup such that only total ordering between binding mode is allowed.
So we must choose whether `/\` or `\/` binds stronger.

ice1000:
@wsx-ucb Yes, and binds stronger. I think we can assume that.

wsx-ucb:
The desired behavior for `/\` and `\/` should be implemented now.
There are still some todos:
- [ ] Look into #515.
- [ ] Add tests for the new behavior. (Not sure how to do this.)

ice1000:
@wsx-ucb For test you can just use assertions like Distiller Test

base/src/main/java/org/aya/distill/BaseDistiller.java
```diff
@@ -124,6 +124,7 @@ private <T extends AyaDocile> Doc visitArg(@NotNull Fmt<T> fmt, @NotNull Arg<T>
   }
 
   public static @NotNull Doc checkParen(@NotNull Outer outer, @NotNull Doc binApp, @NotNull Outer binOp) {
+    if (outer == binOp && (outer == Outer.IMin || outer == Outer.IMax)) return binApp;
```
ice1000, line 127: There's a way to avoid this

wsx-ucb: Oh please let me know!
I admit my solution is a bit hacky.

PR #531, opened 2022-09-21
Allow empty system

ice1000:
@imkiva @lunalunaa is it a good idea to allow `[| i |] A` as a shorthand for `[| i |] A {| |}`?

P.S. I miss GK bc they generate nullability annotations.


imkiva (approved):
Yes it is good as there's no confusion

ice1000:
bors merge

PR #538, opened 2022-09-24
Workaround JDK-8292756

imkiva:
revert this PR when https://bugs.openjdk.org/browse/JDK-8292756 backported to 19


imkiva:
bors merge

PR #545, opened 2022-09-26
Match expression

### LINKED ISSUE #48: Match expression in concrete and core
This should be very simple

wsx-ucb:
Fixes #48.
Current todo list:
- [x] Add match expression in concrete (and syntax?).
- [x] Correctly handle `Match` in various visitors (`Renamer`, `ExprTycker`, etc.).
- [ ] Unify the logic for pattern matching in function definition and that with match expression.
Main difference here is that we can simultaneously match multiple arguments in function definition, but only allowing one scrutinee in the match expression. Maybe just elaborate matching multiple arguments into matching a tuple.
- [x] Add test cases.


ice1000:
Syntax is already there. Just patch the producer. However, I recommend you to wait until we migrate the frontend.

ice1000:
:fire:

ice1000:
ping

wsx-ucb:
Type checking for `match` is implemented by reusing the `elabClausesClassified`.
However, currently not even the simplest case would pass, due to some weird patterns being generated in the match clauses.
I need to reinvestigate all the visitors used after `AyaGKProducer` for this matter.

ice1000:
> Type checking for `match` is implemented by reusing the `elabClausesClassified`.
> However, currently not even the simplest case would pass, due to some weird patterns being generated in the match clauses.
> I need to reinvestigate all the visitors used after `AyaGKProducer` for this matter.

When do you have time? We can debug it together.

wsx-ucb:
I've looked into this issue yesterday, and one problem is that `ExprTraversal` does not visit patterns.
This makes the pattern preprocessing logic applies to function declaration with pattern matching statements, but not to `match` expressions in function body.
The fix wouldn't be hard, but more likely there are other issues hiding around.

ice1000:
> I've looked into this issue yesterday, and one problem is that `ExprTraversal` does not visit patterns.
> This makes the pattern preprocessing logic applies to function declaration with pattern matching statements, but not to `match` expressions in function body.
> The fix wouldn't be hard, but more likely there are other issues hiding around.

I think there is a subinterface of `ExprTraversal` that visits patterns. You should just copy and paste it.

ice1000:
@wsx-ucb from now on, you no longer need to regenerate the lexer. Only the parser. Yay!

base/src/main/java/org/aya/concrete/desugar/Desugarer.java
```diff
@@ -37,14 +37,14 @@ public static class DesugarInterruption extends Exception {}
 
   @Override public @NotNull Expr pre(@NotNull Expr expr) {
     return switch (expr) {
-      case Expr.AppExpr(var pos, Expr.RawSortExpr(var uPos, var kind), var arg)when kind == SortKind.Type -> {
+      case Expr.AppExpr(var pos,Expr.RawSortExpr(var uPos,var kind),var arg)when kind == SortKind.Type -> {
```
ice1000, line 40: Please upgrade IntelliJ

ice1000 (changes requested):

base/src/main/java/org/aya/concrete/desugar/Desugarer.java
wsx-ucb: I'm on 2022.2.3 now, but it seems that the formatting is still wrong.
Which version should I use then?

ice1000: EAP

wsx-ucb: Ok, I've switched to EAP and some spaces are added back.
But there is still no space between `)` and `when`, is this fine at this point?

ice1000:
We should have a meeting and discuss these changes.

base/src/main/java/org/aya/concrete/desugar/Desugarer.java
ice1000: Yes

ice1000 (approved):
bors r+

PR #552, opened 2022-10-04
Expr: Array

HoshinoTented:
- [x] Producer for ARRAY_EXPR
- [x] ExprResolver for Expr.Array
- [x] Desugarer for Expr.Array
- [x] Pretty Print for Expr.Array (I am not sure)

PR #556, opened 2022-10-05
Expr: Array

HoshinoTented:
- [x] Producer for ARRAY_EXPR
- [x] ExprResolver for Expr.Array
- [x] Desugarer for Expr.Array
- [x] Pretty Print for Expr.Array (I am not sure)

#552


base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -154,6 +154,20 @@ yield visitCalls(false, fn, (nc, l) -> l.toDoc(options), outer,
           .map(app -> term(Outer.Free, app)))
       );
       case Expr.Do aDo -> throw new UnsupportedOperationException("TODO");
```
ice1000, line 156: Can you implement this too?

base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
```diff
@@ -116,6 +116,47 @@ public record ExprResolver(
         if (type == path.type() && par == path.partial()) yield path;
         yield new Expr.Path(path.sourcePos(), path.params(), type, par);
       }
+      case Expr.Array arrayExpr -> arrayExpr.arrayBlock().fold(
+        left -> {
+          var bindName = resolve(left.bindName(), ctx);
+          var pureName = resolve(left.pureName(), ctx);
+
+          // region The same coding as Expr.Do! 😭
```
ice1000, line 124: Maybe you can try generalizing the code?

base/src/test/java/org/aya/concrete/DesugarTest.java
```diff
@@ -54,6 +54,16 @@ public class DesugarTest {
       }""");
   }
 
+  @Test public void arrays() {
+    desugarAndPretty("""
+    def simpleList => [ 1, 2, 3 ]
+    def listGenerator => [ x * y | x <- xs, y <- ys ]
```
ice1000, line 60: I guess you need to format this code

base/src/test/java/org/aya/concrete/DesugarTest.java
```diff
@@ -54,6 +54,16 @@ public class DesugarTest {
...
+    """, """
+    def simpleList => :< 1 (:< 2 (:< 3 nil))
+    def listGenerator => >>= xs (\\ (x : _) => >>= ys (\\ (y : _) => pure (x * y)))
+    """);   // Aya doesn't know the `:<` and `>>=` are infix function in this case, how to let she know?
```
ice1000, line 64: I think you have to define those functions. Write something like `def infix >>= => {??}` I guess

ice1000 (commented):
Rest looks good

ice1000:
I also think `desugarAndPretty` is wrongly implemented. It doesn't resolve. Can you fix it in another PR? Like adding resolve and fix those tests who fails by adding unresolved definitions

base/src/test/java/org/aya/concrete/DesugarTest.java
ice1000: No, I think it should be indented

base/src/test/java/org/aya/concrete/DesugarTest.java
```diff
@@ -54,6 +54,27 @@ public class DesugarTest {
...
+      data List (A : Type 0)
+        | nil
+        | :< (_ : A) (_ : List A)
+      open List hiding ()
+      def >>= => {??}
+      def simpleList => :< 1 (:< 2 (:< 3 nil))
+      def listGenerator => >>= xs (\\ (x : _) => >>= ys (\\ (y : _) => pure (x * y)))
+      """); // TODO: make desugarAndPretty does resolve [#556](https://github.com/aya-prover/aya-dev/pull/556#issuecomment-1268376917).
+  }
+
```
imkiva, line 77: I would suggest testing the desugar in some `.aya` files, like the code below that tests the BinOp desugar:

https://github.com/aya-prover/aya-dev/blob/5e317b38ea27dbe878a1f13369102ca94e6044eb/base/src/test/resources/success/src/Assoc.aya#L13-L25

cli/src/main/java/org/aya/cli/parse/AyaGKProducer.java
```diff
@@ -604,13 +604,55 @@ public TeleDecl.FnDecl fnDecl(@NotNull GenericNode<?> node, Stmt.Accessibility a
           })
           .toImmutableSeq());
     }
-    // TODO: implement this
     if (node.is(ARRAY_EXPR)) {
-      return new Expr.HoleExpr(pos, false, null);
+      var arrayBlock = node.peekChild(ARRAY_BLOCK);
+
+      if (arrayBlock != null) {
+        if (arrayBlock.is(ARRAY_COMP_BLOCK)) {
```
imkiva, line 611: - extract two helper methods: `arrayCompBlock(GenericNode)` and `arrayElementList(GenericNode)`
- return early when `arrayBlock == null`

cli/src/main/java/org/aya/cli/parse/AyaGKProducer.java
```diff
@@ -604,13 +604,55 @@ public TeleDecl.FnDecl fnDecl(@NotNull GenericNode<?> node, Stmt.Accessibility a
...
+
     return unreachable(node);
   }
 
+  /**
+   * This function assumed that the node is DO_BINDING
+   */
+  public @NotNull Expr.DoBind buildDoBinding(@NotNull GenericNode<?> node) {
+    var wp = weakId(node.child(WEAK_ID));
+    return new Expr.DoBind(wp.sourcePos(), new LocalVar(wp.data()), expr(node.child(EXPR)));
```
imkiva, line 653: Use `LocalVar.from(wp)`

base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
ice1000: No it's not. Your version is better.

cli/src/main/java/org/aya/cli/parse/AyaGKProducer.java
imkiva: Yes and braces can be omitted like
```
if (arrayBlock == null) return ...;
if (arrayBlock.is(...)) return arrayCompBlock(arrayBlock);
if (arrayBlock.is(...)) return arrayElementList(arrayBlock);

return unreachable(arrayBlock);
```

ice1000 (approved):
Cool!!
bors r+

PR #559, opened 2022-10-06
Pretty: Expr.Do

HoshinoTented:
My idea is that render an `Expr.Do` as Single Line or Multi-Line (see below)

Single Line (all bindings in one line):

```
do { x <- xs, y <- ys, return (x + y) }
```

Multi-Line (one binding each line):

```
do {
  x <- xs,
  y <- ys,
  return (x + y)
}
```

I dislike (simply `Doc.sep` when `lineRemaining` is insufficient):
```
do { x <- xs, y <- ys,
  return (x + y)
}
```

Can we advise the PrettyPrinter to flat something even though the page width is enough?

### DIFF
--- a/base/src/main/java/org/aya/distill/ConcreteDistiller.java
+++ b/base/src/main/java/org/aya/distill/ConcreteDistiller.java
@@ -155,21 +155,17 @@ public class ConcreteDistiller extends BaseDistiller<Expr> {
           .map(app -> term(Outer.Free, app)))
       );
       case Expr.Do doExpr -> {
-        var doBlockDoc = doExpr.binds().map(bind -> {
-          if (bind.var() == LocalVar.IGNORED) {
-            return term(Outer.Free, bind.expr());
-          } else {
-            return visitDoBind(bind);
-          }
-        });
+        var doBlockDoc = doExpr.binds().map(this::visitDoBinding);
 
         // Either not flat (single line) or full flat
-        yield Doc.flatAltBracedBlock(
-          Doc.commaList(doBlockDoc),
-          Doc.vcommaList(
-            doBlockDoc.map(x -> Doc.nest(2, x))
-          )
-        );
+        yield Doc.stickySep(      // doExpr is atom! It cannot be `do\n{ ... }`
+          Doc.symbol("do"),
+          Doc.flatAltBracedBlock(
+            Doc.commaList(doBlockDoc),
+            Doc.vcommaList(
+              doBlockDoc.map(x -> Doc.nest(2, x))
+            )
+        ));
       }
     };
   }
@@ -332,11 +328,13 @@ public class ConcreteDistiller extends BaseDistiller<Expr> {
   /**
    * This function assumed that <code>doBind.var()</code> is not {@link org.aya.ref.LocalVar#IGNORED}
    */
-  public @NotNull Doc visitDoBind(@NotNull Expr.DoBind doBind) {
-    return Doc.sep(
-      varDoc(doBind.var()),
-      Doc.symbol("<-"),
-      term(Outer.Free, doBind.expr()));
+  public @NotNull Doc visitDoBinding(@NotNull Expr.DoBind doBind) {
+    return doBind.var() == LocalVar.IGNORED
+      ? term(Outer.Free, doBind.expr())
+      : Doc.sep(
+          varDoc(doBind.var()),
+          Doc.symbol("<-"),
+          term(Outer.Free, doBind.expr()));
   }
 
   public @NotNull Doc visitPersonality(@NotNull Decl.Personality personality) {
--- a/base/src/main/java/org/aya/distill/ConcreteDistiller.java
+++ b/base/src/main/java/org/aya/distill/ConcreteDistiller.java
@@ -159,7 +159,7 @@ public class ConcreteDistiller extends BaseDistiller<Expr> {
 
         // Either not flat (single line) or full flat
         yield Doc.stickySep(      // doExpr is atom! It cannot be `do\n{ ... }`
-          Doc.symbol("do"),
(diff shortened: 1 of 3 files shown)


base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -153,7 +154,23 @@ yield visitCalls(false, fn, (nc, l) -> l.toDoc(options), outer,
...
+          } else {
+            return visitDoBind(bind);
+          }
+        });
+
+        // Either not flat (single line) or full flat
+        yield Doc.flatAltBracedBlock(
+          Doc.commaList(doBlockDoc),
+          Doc.vcommaList(
+            doBlockDoc.map(x -> Doc.nest(2, x))     // TODO[hoshino]: constant indent?
```
ice1000, line 170: I think this is good for now

pretty/src/main/java/org/aya/pretty/doc/Doc.java
```diff
@@ -242,6 +242,18 @@ record PageWidth(@NotNull IntFunction<Doc> docBuilder) implements Doc {
...
     return wrap("{", "}", doc);
   }
+
+  /**
+   * Either `{ defaultDoc }` or `{\nflatDoc\n}`
+   * TODO: There should be a better impl for this! For example, a new Doc element class
+   */
+  static @NotNull Doc flatAltBracedBlock(Doc defaultDoc, Doc flatDoc) {
+    return flatAlt(
+      stickySep(Doc.plain("{"), defaultDoc, Doc.plain("}")),
```
ice1000, line 252: Use `Doc.symbol("{")`

pretty/src/main/java/org/aya/pretty/doc/Doc.java
```diff
@@ -518,6 +530,10 @@ record PageWidth(@NotNull IntFunction<Doc> docBuilder) implements Doc {
     return join(COMMA, docs);
   }
 
+  @Contract("_ -> new") static @NotNull Doc vcommaList(@NotNull SeqLike<Doc> docs) {
+    return join(cat(plain(","), line()), docs);
```
ice1000, line 534: `plain(",")` -> `COMMA`

pretty/src/main/java/org/aya/pretty/doc/Doc.java
ice1000: If you believe in some backend there will be special treatment for the symbol. For example, in LaTeX backend, `{` should be escaped as `\{`, and this is a special treatment, and hence you should use `symbol("{")`. In general, if you _feel_ like something is not some text, then probably symbol.

pretty/src/main/java/org/aya/pretty/doc/Doc.java
ice1000: I see! Thanks for the explanation.

base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -153,7 +154,23 @@ yield visitCalls(false, fn, (nc, l) -> l.toDoc(options), outer,
         Doc.join(Doc.symbol("|"), idiom.barredApps().view()
           .map(app -> term(Outer.Free, app)))
       );
-      case Expr.Do aDo -> throw new UnsupportedOperationException("TODO");
+      case Expr.Do doExpr -> {
+        var doBlockDoc = doExpr.binds().map(bind -> {
+          if (bind.var() == LocalVar.IGNORED) {
+            return term(Outer.Free, bind.expr());
+          } else {
+            return visitDoBind(bind);
```
imkiva, line 162: We can omit these braces by `doExpr.binds().map(bind -> bind.var() == LocalVar.IGNORED ? term(Outer.Free, bind.expr()) : visitDoBind(bind));`

base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -312,6 +329,16 @@ private Stmt.Accessibility defaultAcc(@NotNull Decl.Personality personality) {
     };
   }
 
+  /**
+   * This function assumed that <code>doBind.var()</code> is not {@link org.aya.ref.LocalVar#IGNORED}
+   */
+  public @NotNull Doc visitDoBind(@NotNull Expr.DoBind doBind) {
```
imkiva, line 335: Put the `LocalVar.IGNORED` check here?

imkiva (commented):
Looks good!

base/src/main/java/org/aya/distill/ConcreteDistiller.java
```diff
@@ -155,21 +155,17 @@ yield visitCalls(false, fn, (nc, l) -> l.toDoc(options), outer,
...
 
         // Either not flat (single line) or full flat
-        yield Doc.flatAltBracedBlock(
-          Doc.commaList(doBlockDoc),
-          Doc.vcommaList(
-            doBlockDoc.map(x -> Doc.nest(2, x))
-          )
-        );
+        yield Doc.stickySep(      // doExpr is atom! It cannot be `do\n{ ... }`
+          Doc.symbol("do"),
```
imkiva, line 162: `Doc.styled(KEYWORD, "do")`

ice1000:
> Where the test should be? I placed them in `ParseTest`, but its name is not good (We have to prevent the doExpr from desugaring)

Ιs there DistillerTest or PrettyTest?

ice1000:
@HoshinoTented I merged your (another) PR and there are now conflicts. Can you resolve them?

ice1000:
I merged your PR into another branch

ice1000:
@HoshinoTented You can make it public

PR #581, opened 2022-10-12
Make LSP Colorful

HoshinoTented:
The goals of this PR are:

+ [ ] Configurable LSP options
+ [ ] Rendering Doc by using LSP options instead of `debugRender()`

### DIFF
--- a/lsp/src/main/java/module-info.java
+++ b/lsp/src/main/java/module-info.java
@@ -15,6 +15,6 @@ module aya.lsp {
 
   opens org.aya.lsp.models to com.google.gson;
   exports org.aya.lsp.actions;
-    exports org.aya.lsp.options;
-    opens org.aya.lsp.options to com.google.gson;
+  exports org.aya.lsp.options;
+  opens org.aya.lsp.options to com.google.gson;
 }
--- a/lsp/src/main/java/module-info.java
+++ b/lsp/src/main/java/module-info.java
@@ -15,4 +15,6 @@ module aya.lsp {
 
   opens org.aya.lsp.models to com.google.gson;
   exports org.aya.lsp.actions;
+    exports org.aya.lsp.options;
+    opens org.aya.lsp.options to com.google.gson;
 }
--- a/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
+++ b/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
@@ -1,0 +1,30 @@
+// Copyright (c) 2020-2022 Tesla (Yinsen) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.lsp.models;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * The server options for LSP, see {@link org.javacs.lsp.InitializeParams#initializationOptions}
+ *
+ * @see RenderOptions
+ */
+public class ServerOptions {
+  /**
+   * the color scheme for pretty printer, can be either {@link String} or {@link JsonObject}
+   * <ul>
+   *   <li>String: color scheme name, "emacs" and "intellij" are valid for now, see {@link org.aya.pretty.style.AyaColorScheme}</li>
+   *   <li>JsonObject: a color scheme string-integer map, see {@link org.aya.pretty.style.AyaColorScheme.Key}</li>
+   * </ul>
+   */
+  public @Nullable JsonElement colorScheme;
+
+  /**
+   * unused for now
+   */
+  public @Nullable JsonElement styleFamily;
+
+  public @Nullable RenderOptions.RenderTarget renderTarget;
+}
--- a/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
+++ b/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
@@ -27,4 +27,10 @@ public class ServerOptions {
   public @Nullable JsonElement styleFamily;
 
   public @Nullable RenderOptions.RenderTarget renderTarget;
+
+  public ServerOptions(@Nullable JsonElement colorScheme, @Nullable JsonElement styleFamily, @Nullable RenderOptions.RenderTarget renderTarget) {
+    this.colorScheme = colorScheme;
+    this.styleFamily = styleFamily;
+    this.renderTarget = renderTarget;
+  }
 }
--- a/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
+++ b/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
@@ -23,7 +23,7 @@ public class ServerOptions {
   public @Nullable Map<String, String> colorOverride;
 
   /**
-   * The colorScheme name, "emacs", "intellij" and "none" are valid for now, see {@link RenderOptions#BUILTIN_COLOR_SCHEMES}.
+   * The colorScheme name, "emacs"/"Emacs" and "intellij"/"IntelliJ" are valid for now, see {@link RenderOptions#BUILTIN_COLOR_SCHEMES}.
    * "none" is picked if {@code colorScheme == null}
    */
   public @Nullable String colorName;
--- a/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
+++ b/lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
@@ -4,28 +4,36 @@ package org.aya.lsp.models;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import org.javacs.lsp.InitializeParams;
 import org.jetbrains.annotations.Nullable;
 
 /**
- * The server options for LSP, see {@link org.javacs.lsp.InitializeParams#initializationOptions}
+ * The server options for LSP
  *
  * @see RenderOptions
+ * @see org.javacs.lsp.InitializeParams#initializationOptions
+ * @see org.aya.lsp.server.AyaLanguageServer#updateOptions(ServerOptions)
  */
 public class ServerOptions {
   /**
    * the color scheme for pretty printer, can be either {@link String} or {@link JsonObject}
    * <ul>
    *   <li>String: color scheme name, "emacs" and "intellij" are valid for now, see {@link org.aya.pretty.style.AyaColorScheme}</li>
-   *   <li>JsonObject: a color scheme string-integer map, see {@link org.aya.pretty.style.AyaColorScheme.Key}</li>
+   *   <li>JsonObject: a {@link String}-{@link Integer} map, see {@link org.aya.pretty.style.AyaColorScheme.Key}</li>
    * </ul>
    */
   public @Nullable JsonElement colorScheme;
 
   /**
    * unused for now
+   *
+   * @see RenderOptions#styleFamily()
    */
   public @Nullable JsonElement styleFamily;
 
+  /**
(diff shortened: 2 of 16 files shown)


lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
```diff
@@ -135,9 +145,28 @@ private void mockLibraries(@NotNull Path path) {
...
+    var options = param.initializationOptions;
+    if (options == null) return;
+
+    try {
+      var opts = new Gson().fromJson(options, ServerOptions.class);
+      updateOptions(opts);
+    } catch (JsonParseException ex) {
+      // TODO[hoshino]: warn or panic?
+      // Do warn
+      Log.e("Unable to convert an InitializeParams.initializationOptions to a ServerOptions, cause: %s", ex.getMessage());
```
imkiva, line 166: A warning is enough here since `RenderOptions` is an aya-specific feature.

lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
```diff
@@ -376,6 +405,26 @@ public ComputeTermResult computeTerm(@NotNull ComputeTermResult.Params input, Co
     return primFactories.getOrPut(owner.underlyingLibrary(), LspPrimFactory::new);
   }
 
+  @LspRequest("options/updateOptions")
+  public void updateOptions(@NotNull ServerOptions options) {
```
imkiva, line 409: Prefix the method name with `aya/` in case LSP spec adds some methods prefixed with `options`

lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
```diff
@@ -376,6 +405,26 @@ public ComputeTermResult computeTerm(@NotNull ComputeTermResult.Params input, Co
...
+    } else {
+      this.renderOptions.getAndUpdate(opts -> opts.update(renderOpts.get()));
+    }
+  }
+
+  /**
+   * A test only API, throws an InternalException
+   */
+  @Contract("-> fail")
+  @LspRequest("internal/panic")
```
imkiva, line 423: Can we move this to some `AyaTestLanguageServer` ?

pretty/src/main/java/org/aya/pretty/style/AyaStyleFamily.java
```diff
@@ -13,26 +13,47 @@
...
-    Tuple.of("aya:Keyword", Style.bold().and().color("aya:Keyword")),
-    Tuple.of("aya:PrimCall", Style.color("aya:Keyword").and()),
-    Tuple.of("aya:FnCall", Style.color("aya:FnCall").and()),
-    Tuple.of("aya:DataCall", Style.color("aya:DataCall").and()),
-    Tuple.of("aya:StructCall", Style.color("aya:StructCall").and()),
-    Tuple.of("aya:ConCall", Style.color("aya:ConCall").and()),
-    Tuple.of("aya:FieldCall", Style.color("aya:FieldCall").and()),
-    Tuple.of("aya:Generalized", Style.color("aya:Generalized").and())
+    Tuple.of(Key.Keyword.key(), Style.bold().and().color("aya:Keyword")),
+    Tuple.of(Key.PrimCall.key(), Style.color("aya:Keyword").and()),
```
imkiva, line 39:
```suggestion
    Tuple.of(Key.Keyword.key(), Style.bold().and().color(AyaColorScheme.Key.Keyword.key())),
    Tuple.of(Key.PrimCall.key(), Style.color(AyaColorScheme.Key.Keyword.key()).and()),
```

lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class ServerOptions {
+  /**
+   * the color scheme for pretty printer, can be either {@link String} or {@link JsonObject}
+   * <ul>
+   *   <li>String: color scheme name, "emacs" and "intellij" are valid for now, see {@link org.aya.pretty.style.AyaColorScheme}</li>
+   *   <li>JsonObject: a {@link String}-{@link Integer} map, see {@link org.aya.pretty.style.AyaColorScheme.Key}</li>
+   * </ul>
+   */
+  public @Nullable JsonElement colorScheme;
```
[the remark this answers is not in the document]
imkiva: > I preferred the all-undefined ColorScheme as the default one.

Yes

lsp/src/main/java/org/aya/lsp/options/RenderOptions.java
```diff
@@ -0,0 +1,198 @@
...
+  private static Result<ColorScheme, String> parseColorScheme(@NotNull JsonElement colorScheme) {
+    if (colorScheme.isJsonPrimitive() && colorScheme.getAsJsonPrimitive().isString()) {
+      var colorSchemeName = colorScheme.getAsString();
+
+      if (colorSchemeName.equals("emacs")) {
+        return Result.ok(AyaColorScheme.EMACS);
+      } else if (colorSchemeName.equals("intellij")) {
+        return Result.ok(AyaColorScheme.INTELLIJ);
+      } else {
+        return Result.err("Invalid color scheme name");
```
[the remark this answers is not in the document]
imkiva: > Maybe a String(Name)-ColorScheme map.

Exactly!

lsp/src/main/java/org/aya/lsp/options/RenderOptions.java
```diff
@@ -0,0 +1,216 @@
...
+
+      var isKeyValid = key != null && Seq.from(AyaColorScheme.Key.values()).view()
+        .map(AyaColorScheme.Key::key)
+        .contains(key);
+      var isValueValid = value != null;   // I don't like to inline this code, sorry!
+
+      if (!isKeyValid) return Result.err("Invalid key: " + key);
+      if (!isValueValid) return Result.err("Invalid color: null");
+
+      var colorResult = parseColor(value);
```
imkiva, line 105: Just ignore them instead of aborting the whole parser for future compatibility. The code could also be simplfiied in this case.

lsp/src/main/java/org/aya/lsp/options/RenderOptions.java
```diff
@@ -0,0 +1,216 @@
...
+      var color = colorResult.get();
+      var oldValue = newColorScheme.put(key, color);
+
+      if (oldValue.isDefined() && ! oldValue.get().equals(color)) {
+        modified = true;
+      }
+    }
+
+    return Result.ok(
+      modified ? new AyaColorScheme(newColorScheme) : base);
```
imkiva, line 117: Any rationale for reusing the `base`? They are stateless so using a copy does not break or make anything.
I suggest always making a copy which makes the parsing code here simplified like
```java
base.definiedColors().forEach((k, v) -> newColor.put(k,
  override.getOption(k).flatMap(this::parseColor).getOrDefault(v)))
  // this::parseColor : String -> Result<Integer, Err> or something alike
```

lsp/src/main/java/org/aya/lsp/options/RenderOptions.java
```diff
@@ -0,0 +1,216 @@
...
+   */
+  private static Result<ColorScheme, String> parseColorScheme(@NotNull ServerOptions options) {
+    var colorName = options.colorName;
+    var baseOption = options.colorName == null
+      ? Option.some(AyaColorScheme.EMPTY)
+      : BUILTIN_COLOR_SCHEMES.getOption(colorName);
+
+    if (baseOption.isEmpty()) {
+      return Result.err("Invalid color scheme name");
+    }
```
imkiva, line 83: When the client sets `colorName` to any invalid name, say `invalid`, we will throw this error instead of using `EMPTY`. Is this intentional? I suggest removing `none` from `BUILTIN_COLOR_SCHEMES` and using
```java
var baseOption = BUILTIN_COLOR_SCHEMES.getOrDefault(colorName, AyaColorScheme.EMPTY);
```

Because the empty color scheme should be a fallback rather than an option.

lsp/src/main/java/org/aya/lsp/options/RenderOptions.java
```diff
@@ -0,0 +1,216 @@
...
+
+  /**
+   * Construct a {@link RenderOptions} from {@link ServerOptions}
+   */
+  @Contract(pure = true)
+  public static @NotNull Result<@NotNull RenderOptions, @NotNull String> fromServerOptions(@NotNull ServerOptions options) {
+    Option<ColorScheme> colorScheme = Option.none();
+
+    // colorScheme is none if both colorName and colorOverride are null
+    if (options.colorName != null || options.colorOverride != null) {
```
imkiva, line 60: I see `parseColorScheme` produces `EMPTY` when `colorName` is not set. Then it should be the same here --- `colorName` is not set. 

In particular, use `EMPTY` here so:
- the field `RenderOptions#colorScheme` does not need to live in an `Option`
- some `getOrDefault` checks in `renderXXX` can be deleted.

If we do the same to `RenderOptions#styleFamily`, the `RenderOptions.DEFAULT` can be deleted.

lsp/src/main/java/org/aya/lsp/options/RenderOptions.java
```diff
@@ -0,0 +1,216 @@
...
+   * @param capabilities the client capabilities, `plaintext` is returned if the capabilities is empty.
+   * @throws IllegalArgumentException is thrown if the capabilities is not empty and invalid.
+   */
+  public static @NotNull RenderTarget renderTargetFromCapabilities(@NotNull List<String> capabilities)
+    throws IllegalArgumentException {
+    if (capabilities.isEmpty()) return RenderTarget.Debug;
+    if (capabilities.contains(MarkupKind.Markdown)) return RenderTarget.HTML;
+    if (capabilities.contains(MarkupKind.PlainText)) return RenderTarget.Debug;
+
+    throw new IllegalArgumentException("Unsupported capabilities: " + capabilities);
```
imkiva, line 169: Should not throw here, instead, return `RenderTarget.Debug` because the client should at least support plain text format.

If you want to tell the client "I don't understand", you should do it in LSP initialization (even in that case you should only raise a warning) where the capabilities are received.

lsp/src/main/java/module-info.java
```diff
@@ -15,4 +15,6 @@
 
   opens org.aya.lsp.models to com.google.gson;
   exports org.aya.lsp.actions;
+    exports org.aya.lsp.options;
+    opens org.aya.lsp.options to com.google.gson;
```
imkiva, line 19: I think you need a `Ctrl - Alt - L`

imkiva (commented):
Some ideas on when to raise an error or a warning:

- If something is mandatory, raise an error when we failed to achieve that.
- If something is optional, or just an enhancement on a specific feature and there's at least one fallback, raise a warning and choose the fallback.

lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
imkiva: > `aya/options/updateOptions` or `aya/updateOptions` ?

Choose the one you like most 😉

lsp/src/main/java/org/aya/lsp/models/ServerOptions.java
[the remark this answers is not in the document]
imkiva: But why not let clients send the same color scheme again when they only want to update the style family? 
And what if the clients want to clear the color scheme? -- A none is always ignored so clients can't do so.

ice1000:
Any hope of splitting into smaller parts?

PR #583, opened 2022-10-15
Fix #528 by not using unification

### LINKED ISSUE #528: Why not use unification
https://github.com/aya-prover/aya-dev/blob/25089b867fc15648d82259e39a7b143f739bff2c/base/src/main/java/org/aya/generic/Shaped.java#L33-L56

@imkiva

ice1000:
But instead use normalization

bors d=imkiva


imkiva (approved):
bors  merge

PR #586, opened 2022-10-17
Make PrettyPrinter Alive

HoshinoTented:
https://github.com/aya-prover/aya-dev/pull/581#issuecomment-1279626774


ice1000:
You can do in a separate pr

PR #589, opened 2022-10-19
repr: List

HoshinoTented:
- [x] `Expr.Array` -> `LitTerm.ShapedList`
- [ ] ~~`Pattern.List` -> `Pat.ShapedList`~~
- [x] `LitTerm.ShapedList` ser/deser
- [ ] ~~`Pat.ShapedList` ser/deser~~
- [x] Desugar `Pattern.List` with `CodeShape`

~~I think the code is buggy (and not elegant) for [now](https://github.com/aya-prover/aya-dev/tree/4802bad65280ca341125a92b572035688954a9cc), but they passed the tests.~~ I also need some help with the TODOs.

Edit (by @ice1000): there are two missing things in this PR, but it looks pretty good now.

+ [ ] In case `cons` constructor has implicit arguments, we should generate the correct implicitness accordingly.
+ [ ] I think the infrastructure of `Shaped` can be improved. Currently it's a bit wacky as we use `state` from time to time and it is nullable. I think we can do this: there is a thing "unknown type literal" that bans `constructorForm` and the type is a hole, and "known type literal" which does not need a `state` in `constructorForm` and the type is `DataCall`. When we need to use `constructorForm` of "unknown type literal" in a type-checking function, we try to convert it into known type literal by using the `state` and in this case, it's always NotNull. If it fails we can report error. This makes more operations type-safe and less error-prone.

### DIFF
--- a/base/src/main/java/org/aya/distill/CoreDistiller.java
+++ b/base/src/main/java/org/aya/distill/CoreDistiller.java
@@ -159,6 +159,17 @@ public class CoreDistiller extends BaseDistiller<Term> {
           ? linkLit(0, zero.ref, CON_CALL)
           : linkLit(shaped.repr(), suc.ref, CON_CALL),
         () -> Doc.plain(String.valueOf(shaped.repr())));
+      case LitTerm.ShapedList shaped -> {
+        var display = Doc.sep(
+          Doc.symbol("["),
+          Doc.commaList(shaped.repr().map(x -> term(Outer.Free, x))),
+          Doc.symbol("]"));
+
+        yield shaped.with((nil, cons, dataArg) -> {
+          // link to data, i think.
+          return linkListLit(display, nil.dataRef, DATA_CALL);   // Maybe we should use other style.
+        }, () -> display);
+      }
       case PrimTerm.Str str -> Doc.plain("\"" + StringUtil.escapeStringCharacters(str.string()) + "\"");
       case FormTerm.PartTy ty -> checkParen(outer, Doc.sep(Doc.styled(KEYWORD, "Partial"),
         term(Outer.AppSpine, ty.type()), Doc.parened(restr(options, ty.restr()))), Outer.AppSpine);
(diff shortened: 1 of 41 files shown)


base/src/main/java/org/aya/distill/CoreDistiller.java
```diff
@@ -159,6 +159,17 @@ yield visitCalls(true, defVar(var),
...
         () -> Doc.plain(String.valueOf(shaped.repr())));
+      case LitTerm.ShapedList shaped -> {
+        var display = Doc.sep(
+          Doc.symbol("["),
+          Doc.commaList(shaped.repr().map(x -> term(Outer.Free, x))),
+          Doc.symbol("]"));
+
+        yield shaped.with((nil, cons, dataArg) -> {
+          // link to data, i think.
+          return linkListLit(display, nil.dataRef, DATA_CALL);   // Maybe we should use other style.
```
imkiva, line 170: Maybe we can link `[` to `cons` and `]` to `nil`, which looks funny!

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -229,6 +230,35 @@ public final class ExprTycker extends Tycker {
...
+          new Arg<>(hole._1, dataParam.explicit())));
+
+        // do type check
+        var wellTypeds = MutableArrayList.<Term>create(elements.size());
+
+        for (var element : elements) {
+          var result = inherit(element, hole._1);
+
+          if (result.wellTyped() instanceof ErrorTerm || result.type() instanceof ErrorTerm) yield result;
+          wellTypeds.append(result.wellTyped());
```
imkiva, line 257: It's ok to push `ErrorTerm`s to that list and typecheck the remaining elements.

HoshinoTented: Yes, the `fail` will report the error.

base/src/main/java/org/aya/distill/CoreDistiller.java
HoshinoTented: How about linking `,` to `cons`, `[` and `]` to `nil` ?

ice1000: > How about linking `,` to `cons`, `[` and `]` to `nil` ?

I _love_ this idea

ice1000:
>  Pattern.List -> Pat.ShapedList

This might be unnecessary

ice1000 (approved):
bors r+

Issue #598, opened 2022-10-23
Leftover of #589


ice1000:
+ [x] In case `cons` constructor has implicit arguments, we should generate the correct implicitness accordingly. I think @HoshinoTented can handle this part.
+ [ ] I think the infrastructure of `Shaped` can be improved. Currently it's a bit wacky as we use `state` from time to time and it is nullable. I think we can do this: there is a thing "unknown type literal" that bans `constructorForm` and the type is a hole, and "known type literal" which does not need a `state` in `constructorForm` and the type is `DataCall`. When we need to use `constructorForm` of "unknown type literal" in a type-checking function, we try to convert it into known type literal by using the `state` and in this case, it's always NotNull. If it fails we can report error. This makes more operations type-safe and less error-prone. I think this should go to @imkiva 

See #589

PR #599, opened 2022-10-23
Issue #598, Part 1.

HoshinoTented:
## This PR should:

+ [x] Supports any licit ParamShape (whatever explicit or implicit)
+ [ ] Desugar `[ {O} ]` ~~correctly~~ (or reject it).
+ [ ] Handle `cons` with implicit arguments correctly.

## Questions

+ Should we support `[ {expr/pattern} ]` ?
  - We do not support
    * When do we reject this expression/pattern? Parsing (in `.bnf`)? Producing (in `AyaGKProducer`)?
  - ~~We do support~~
    * ~~How to handle `[ (S n) as m ]` where `| cons {A} (List A)`~~
    * ~~Similarly, how to handle `[ {S n} as m ]` where `| cons (A) (List A)`~~

### DIFF
--- a/base/src/main/java/org/aya/core/repr/CodeShape.java
+++ b/base/src/main/java/org/aya/core/repr/CodeShape.java
@@ -44,7 +44,7 @@ public sealed interface CodeShape {
 
     /**
      * @param superLevel the data def reference
-     * @param args corresponds to {@link CallTerm.Data#args()}
+     * @param args       corresponds to {@link CallTerm.Data#args()}
      */
     record Call(int superLevel, @NotNull ImmutableSeq<TermShape> args) implements TermShape {
       @Contract("_ -> new") public static @NotNull Call justCall(int superLevel) {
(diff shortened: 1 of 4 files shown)


ice1000:
>  Desugar [ {O} ] correctly (or reject it).

I propose to reject it

base/src/main/java/org/aya/core/repr/CodeShape.java
```diff
@@ -75,18 +75,29 @@ record Any() implements PatShape {}
    * @author kiva
    */
   sealed interface ParamShape {
+    // TODO[hoshino]: a better name is requested!
+    enum Explicit {
```
ice1000, line 79: Licit

base/src/main/java/org/aya/core/repr/CodeShape.java
```diff
@@ -75,18 +75,29 @@ record Any() implements PatShape {}
...
+      Any,
+      Explicit,
+      Implicit
```
ice1000, line 82: `Exp`, `Imp`

ice1000:
Alternative: `Explicit.Yes`, Explicit.No`, `Explicit.Unsure`

HoshinoTented:
The base made a big change, close.

ice1000:
RIP

PR #600, opened 2022-10-23
Computation for `coe`

lunalunaa:
This pull request implements the computation rule for `coe` on a Pi type.

### DIFF
--- a/base/src/main/java/org/aya/core/visitor/BetaExpander.java
+++ b/base/src/main/java/org/aya/core/visitor/BetaExpander.java
@@ -2,7 +2,6 @@
 // Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
 package org.aya.core.visitor;
 
-import kala.collection.immutable.ImmutableSeq;
 import org.aya.core.term.*;
 import org.aya.generic.Arg;
 import org.aya.guest0x0.cubical.Partial;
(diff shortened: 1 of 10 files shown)


lunalunaa:
Some further tests are needed.

base/src/main/java/org/aya/core/visitor/BetaExpander.java
```diff
@@ -56,10 +60,89 @@ yield switch (partial(app.cube().partial())) {
...
           yield new IntroTerm.Lambda(param, new RefTerm(var));
         }
-        // TODO: coe computation
-        yield coe;
+
+        var varI = new LocalVar("i");
+        var codom = apply(new ElimTerm.App(coe.type(), new Arg<>(new RefTerm(varI), true)));
+
+        yield switch (codom) {
+          case FormTerm.Path path -> throw new InternalException("TODO");
```
ice1000, line 68: You can yield coe jn this case, as nothing happened

base/src/main/java/org/aya/core/visitor/BetaExpander.java
```diff
@@ -56,10 +60,89 @@ yield switch (partial(app.cube().partial())) {
...
+            var vType = apply(new ElimTerm.App(A, new Arg<>(PrimTerm.Mula.RIGHT, true)));
+            var w = coeFillInv(A, coe.restr(), new RefTerm(vVar));
+            var BSubsted = B.subst(pi.param().ref(), w);
+            var wSusted = w.subst(varI, PrimTerm.Mula.LEFT);
+            yield new IntroTerm.Lambda(new Term.Param(u0Var, new ElimTerm.App(coe.type(), new Arg<>(PrimTerm.Mula.LEFT, true)), true),
+              new IntroTerm.Lambda(new Term.Param(vVar, vType, true),
+                new ElimTerm.App(new PrimTerm.Coe(BSubsted, coe.restr()),
+                  new Arg<>(new ElimTerm.App(new RefTerm(u0Var), new Arg<>(wSusted, true)), true))));
+          }
+          case FormTerm.Sigma sigma -> throw new InternalException("TODO");
```
ice1000, line 83: Same

base/src/main/java/org/aya/core/visitor/BetaExpander.java
```diff
@@ -56,10 +60,89 @@ yield switch (partial(app.cube().partial())) {
...
+          case FormTerm.Type type -> throw new InternalException("TODO");
```
ice1000, line 84: Return id

ice1000:
Congradulations! You opened #600

ice1000 (commented):
I think the current design of `hcomp` needs some changes.

1. Currently, you have `HComp(A, u, phi, u0) : A`. I think it's better to have `HComp(A, phi, u) : A -> A`. This makes the core simpler, and slightly complicates type synthesis.
2. Same code as above -- the type of `u` depends on `phi` so I recommend putting `phi` before `u`.
3. I think you are a bit aggressive on the reduction of `coe`. You don't have to throw exception -- just leave the redex there should also work.

ice1000:
Ooops!

image

So it's not a design problem, but a bug!

lunalunaa:
> I think the current design of `hcomp` needs some changes.
> 
>     1. Currently, you have `HComp(A, u, phi, u0) : A`. I think it's better to have `HComp(A, phi, u) : A -> A`. This makes the core simpler, and slightly complicates type synthesis.
> 
>     2. Same code as above -- the type of `u` depends on `phi` so I recommend putting `phi` before `u`.
> 
>     3. I think you are a bit aggressive on the reduction of `coe`. You don't have to throw exception -- just leave the redex there should also work.

About 1,  We need to check the cubical subtype condition $\Gamma\vdash u_0: A[\phi \mapsto u\ 0]$.

ice1000:
Oh I see. :sob:

PR #606, opened 2022-10-25
Overhaul the `Sub` system of unification

ice1000:
As suggested by Andras Kovacs


ice1000 (commented):
bors r+

Waiting for @Glavo to implement zipView3 in Kala

PR #610, opened 2022-10-26
Issue 609

HoshinoTented:
This PR:

- delete `Pat#toExpr` ~~(unused for now, but it is safe to delete)~~
- insert implicit argument when `PatToTerm`
- improve replacement when `PatTyck` (see `ExprTycker#lets`)


base/src/main/java/org/aya/tyck/pat/PatTycker.java
```diff
@@ -374,9 +399,18 @@ private record PatData(
     tracing(TreeBuilder::reduce);
     addPatSubst(data.param.ref(), res, pat.sourcePos());
     data.results.append(res);
+
+    // update signature for current parameter,
+    // because it probably changed when we call `data.sig.inst`
+    exprTycker.localCtx.put(data.param().ref(), type);
```
ice1000, line 405: I suspect that the extra bindings come from here.

HoshinoTented: of course

ice1000:
https://github.com/aya-prover/aya-dev/blob/c7c57e4ee13577999163d0ef06864f829d7ae7cf/.github/workflows/commit-check.sh#L12

ice1000:
I suggest `pat: improve code` but it's up to you

HoshinoTented:
> I suggest `pat: improve code` but it's up to you

solved

base/src/main/java/org/aya/tyck/pat/PatTycker.java
HoshinoTented: Don't resolve, this is a bug.

ice1000: I'm taking care of it

HoshinoTented:
We have to display all var in `ExprTycker#lets` for users when context information is requested.

HoshinoTented:
We didn't inline any `Term.MetaPat` in the types of Pat, I don't know if it makes any problem.

ice1000:
@HoshinoTented 应该 inline 的

HoshinoTented:
I found that this code won't compile

```
open data Foo (A : Type)
| B => foo A

def bar {A : Type} (x : Foo A) : A
| foo y => y
```

because we didn't subst the A to B. It seems like we need another `checkRhs` for Ctor, but this should be completed in another PR.

HoshinoTented:
More extreme tests are NEEDED!

ice1000:
codecov r+

PR #616, opened 2022-10-28
Computation rule of coe for sigma type

lunalunaa:
This implements computation rule of coe for binary sigma types with a test.


ice1000 (approved):
bors r+

PR #623, opened 2022-11-01
No more useless subst

HoshinoTented:
The _subst of parameters of Ctor_ now disappear after tyck that `Pat.Ctor`. For example:

```aya
open data Nat
| O
| S (_1 : Nat)

def justMatch (m n : Nat) : Nat
| O, n => O
| S m', n => m'
```

There is a subst `_1 |-> m'` when we tyck the  `n` pattern in `| S m', n => m'`, which is useless.

### DIFF
--- a/base/src/main/java/org/aya/tyck/pat/PatTycker.java
+++ b/base/src/main/java/org/aya/tyck/pat/PatTycker.java
@@ -333,8 +333,6 @@ public final class PatTycker {
   public @NotNull Tuple2<SeqView<Pat>, Term>
   visitPatterns(@NotNull Def.Signature sig, @NotNull SeqView<Pattern> stream, @Nullable Pattern outerPattern) {
     var results = MutableList.<Pat>create();
-    // this path seems useless
-    if (sig.param().isEmpty() && stream.isEmpty()) return done(results, sig.result());
     // last pattern which user given (not aya generated)
     @Nullable Pattern lastPat = null;
     while (sig.param().isNotEmpty()) {
--- a/base/src/main/java/org/aya/tyck/pat/PatTycker.java
+++ b/base/src/main/java/org/aya/tyck/pat/PatTycker.java
@@ -59,10 +59,17 @@ public final class PatTycker {
 
   /**
    * An {@code as pattern} map.
-   * Every binding in the function telescope was treated as an {@code as pattern}
-   * so they were added to this map.
    */
   private final @NotNull TypedSubst patSubst;
+
+  /**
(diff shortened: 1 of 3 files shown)


base/src/main/java/org/aya/tyck/pat/PatTycker.java
```diff
@@ -59,21 +59,30 @@ public final class PatTycker {
...
    * An {@code as pattern} map.
-   * Every binding in the function telescope was treated as an {@code as pattern}
-   * so they were added to this map.
    */
   private final @NotNull TypedSubst patSubst;
+
+  /**
+   * Every binding in the function telescope was treated as an {@code as pattern},
+   * but they won't be added to {@link PatTycker#patSubst},
+   * because we may visit a complete different signature during tycking {@link Pat.Ctor},
```
ice1000, line 68: completely

base/src/main/java/org/aya/tyck/pat/PatTycker.java
```diff
@@ -59,21 +59,30 @@ public final class PatTycker {
...
+   * and the substs for that signature become useless after tycking {@link Pat.Ctor}
+   * (Also, the substs for the function we tycking is useless when we tycking the {@link Pat.Ctor}).
```
ice1000, line 70: the function we are tycking... when we tyck the...

base/src/main/java/org/aya/tyck/pat/PatTycker.java
```diff
@@ -320,7 +333,8 @@ yield doTyck(new Pattern.FakeShapedList(
   public @NotNull Tuple2<SeqView<Pat>, Term>
   visitPatterns(@NotNull Def.Signature sig, @NotNull SeqView<Pattern> stream, @Nullable Pattern outerPattern) {
     var results = MutableList.<Pat>create();
-    if (sig.param().isEmpty() && stream.isEmpty()) return Tuple.of(results.view(), sig.result());
+    // this path seems useless
```
ice1000, line 336: Is it still reachable?

ice1000: I mean like does the code coverage shows hit for this line

HoshinoTented: It is reachable, but the behavior won't change if we delete that line:

* `sig.param().isEmpty`: won't go through the `while` loop
* `stream.isEmpty()`: won't go through the `if` test

ice1000: Oh

PR #628, opened 2022-11-03
Don't normalize too early

HoshinoTented:
Fixed: normalizing too early when the pat is binding.

### DIFF
--- a/base/src/main/java/org/aya/core/pat/PatMatcher.java
+++ b/base/src/main/java/org/aya/core/pat/PatMatcher.java
@@ -59,11 +59,11 @@ public record PatMatcher(@NotNull Subst subst, @Nullable LocalCtx localCtx, @Not
    * @throws Mismatch if mismatch
    */
   private void match(@NotNull Pat pat, @NotNull Term term) throws Mismatch {
-    term = pre.apply(term);
     switch (pat) {
       case Pat.Bind bind -> subst.addDirectly(bind.bind(), term);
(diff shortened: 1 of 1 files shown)


wsx-ucb (approved):
LGTM!

base/src/main/java/org/aya/core/pat/PatMatcher.java
```diff
@@ -77,23 +71,22 @@ private void match(@NotNull Pat pat, @NotNull Term term) throws Mismatch {
...
           case RefTerm.MetaPat metaPat -> solve(pat, metaPat);
           default -> throw new Mismatch(true);
         }
       }
-      case Pat.Meta meta -> {
-        var sol = meta.solution().get();
-        assert sol != null : "Unsolved pattern " + meta;
-        match(sol, term);
-      }
+      case Pat.Meta ignored -> throw new InternalException("Pat.Meta is not allowed");
```
imkiva, line 81: Why we do not match the solution against the term? I guess we inlined these solutions before the match?

imkiva (approved):
LGTM and some naive questions (sorry!)

base/src/main/java/org/aya/core/pat/PatMatcher.java
HoshinoTented: I think we should inline all `Pat.Meta` before it can be used to match, so any `Pat.Meta` implies something was not inlined which is a bug.

PR #629, opened 2022-11-04
Implement Prop pattern matching tyck

tsao-chi:
TODO:
+ Generate ErasedTerm for pattern matching

### DIFF
--- a/base/src/main/java/org/aya/tyck/pat/PatTycker.java
+++ b/base/src/main/java/org/aya/tyck/pat/PatTycker.java
@@ -160,7 +160,10 @@ public final class PatTycker {
   }
 
   private LhsResult checkLhs(Pattern.Clause match, Def.Signature signature) {
-    var resultIsProp = signature.result().computeType(exprTycker.state, exprTycker.localCtx) instanceof FormTerm.Prop;
+    var resultIsProp = false;
+    try {
+      resultIsProp = signature.result().computeType(exprTycker.state, exprTycker.localCtx) instanceof FormTerm.Prop;
+    } catch (org.aya.generic.util.InternalException ignored) {} // TODO: remove this hack
     var parent = exprTycker.localCtx;
     exprTycker.localCtx = parent.deriveMap();
     currentClause = match;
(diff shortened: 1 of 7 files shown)


base/src/main/java/org/aya/tyck/pat/PatTycker.java
```diff
@@ -169,10 +169,12 @@ public record LhsResult(
   }
 
   private LhsResult checkLhs(Pattern.Clause match, Def.Signature signature) {
+    var ctx = exprTycker.localCtx.deriveMap();
+    var resultIsProp = ctx.with((() -> signature.result().computeType(exprTycker.state, ctx) instanceof FormTerm.Prop), signature.param().toArray(new Term.Param[0]));
     var parent = exprTycker.localCtx;
     exprTycker.localCtx = parent.deriveMap();
```
HoshinoTented, line 175: `ctx` looks duplicated with `parent.deriveMap()` / `exprTycker.localCtx` after line 175, How about:

```java
var parent = exprTycker.localCtx;
exprTycker.localCtx = parent.deriveMap();
exprTycker.localCtx.with(...);
```

or we let `LocalCtx#with` derive a `LocalCtx` automatically.

HoshinoTented (changes requested):

base/src/main/java/org/aya/tyck/pat/PatTycker.java
tsao-chi: I like the second solution

tsao-chi: We might address it in future prs

HoshinoTented (dismissed):

base/src/main/java/org/aya/tyck/pat/PatTycker.java
ice1000: I think this is sufficiently simple for this PR

ice1000 (commented):
I'll try to change _something_

PR #633, opened 2022-11-05
Disallow using `coe` as function

imkiva:
this is done by rejecting references to `coe` in the resolver.

### DIFF
--- a/base/src/main/java/org/aya/core/def/PrimDef.java
+++ b/base/src/main/java/org/aya/core/def/PrimDef.java
@@ -80,7 +80,7 @@ public final class PrimDef extends TopLevelDef<Term> {
 
   public static class Factory {
     private final class Initializer {
-      public final @NotNull PrimDef.PrimSeed coerce = new PrimSeed(ID.COE, this::coe, ref -> {
+      public final @NotNull PrimDef.PrimSeed coe = new PrimSeed(ID.COE, this::coe, ref -> {
         var varA = new LocalVar("A");
         var paramA = new Term.Param(varA, intervalToA(), true);
         var paramRestr = new Term.Param(new LocalVar("i"), IntervalTerm.INSTANCE, true);
@@ -217,7 +217,7 @@ public final class PrimDef extends TopLevelDef<Term> {
           init.stringConcat,
           init.intervalType,
           init.partialType,
-          init.coerce,
+          init.coe,
           init.hcomp
         ).map(seed -> Tuple.of(seed.name, seed))
         .toImmutableMap();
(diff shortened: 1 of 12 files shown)


base/src/test/resources/success/src/Refparo/Model.aya
```diff
@@ -4,10 +4,10 @@ public open data Unit | unit
 
 -- we cannot eta-contract this, same problem as Cubical Agda
 def coe {A B : Type} (eq : A ↑ = B) (x : A) : B
-  => ↑ primCoe (\y => eq y) 0 x
+  => (\y => eq y).primCoe x
 
 def coeFill {A : Type} {x : A} : x = coe (↑ idp) x
-  => λ i => ↑ primCoe (λ j => A) (~ i) x
+  => λ i => (λ j => A).primCoe x freeze (~ i)
```
imkiva, line 10: Is this a bug?

ice1000: image

base/src/main/java/org/aya/core/def/PrimDef.java
```diff
@@ -217,7 +217,7 @@ public Factory() {
           init.stringConcat,
           init.intervalType,
           init.partialType,
-          init.coerce,
+          init.coe,
```
ice1000, line 220: Tbh I personally prefer longer names in the compiler...

base/src/main/java/org/aya/resolve/error/PrimResolveError.java
```diff
@@ -47,9 +47,25 @@ record Dependency(
...
   }
+
+  record BadUsage(
+    @NotNull String name,
+    @NotNull SourcePos sourcePos
+  ) implements PrimResolveError {
+    @Override public @NotNull Doc describe(@NotNull DistillerOptions options) {
+      return Doc.sep(Doc.english("The primitive"),
+        Doc.styled(Style.code(), Doc.plain(name)),
+        Doc.english("is not designed to be used as a function"));
```
ice1000, line 63: Maybe suggest the proper way to use it?

ice1000 (commented):
rua!

base/src/main/java/org/aya/resolve/error/PrimResolveError.java
imkiva: See the `hint` function below

base/src/main/java/org/aya/core/def/PrimDef.java
imkiva: Okay, but for `coerce` I guess it's fine, because I decide to name `coeInv` as `eoc` so `coerce` looks wired as `ecreoe`

base/src/test/resources/success/src/Refparo/Model.aya
imkiva: I still think it is abnormal... I will try giving a counterexample without using primitives.

ice1000: :thinking:

PR #634, opened 2022-11-06
Allowing Implicit Argument in List Ctor

HoshinoTented:
A re-development of #599 on the new base.

### DIFF
--- a/base/src/test/java/org/aya/repr/ShapeMatcherTest.java
+++ b/base/src/test/java/org/aya/repr/ShapeMatcherTest.java
@@ -52,6 +52,17 @@ public class ShapeMatcherTest {
       """);
   }
 
+  @Test
+  public void matchWeirdList() {
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List {A : Type} | nil | cons A (List {A})");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List (A : Type) | nil | cons {A} (List A)");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List (A : Type) | nil | cons A {List A}");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List {A : Type} | nil | cons {A} (List {A})");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List {A : Type} | nil | cons A {List {A}}");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List (A : Type) | nil | cons {A} {List A}");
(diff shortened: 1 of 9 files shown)


base/src/test/java/org/aya/repr/ShapeMatcherTest.java
```diff
@@ -52,6 +52,17 @@ data List (A : Type)
...
+  @Test
+  public void matchWeirdList() {
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List {A : Type} | nil | cons A (List {A})");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List (A : Type) | nil | cons {A} (List A)");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List (A : Type) | nil | cons A {List A}");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List {A : Type} | nil | cons {A} (List {A})");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List {A : Type} | nil | cons A {List {A}}");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List (A : Type) | nil | cons {A} {List A}");
+    match(true, AyaShape.AyaListShape.DATA_LIST, "data List {A : Type} | nil | cons {A} {List {A}}");
+  }
```
imkiva, line 64: Should we support them? The code shapes are designed to optimize common data structures, while these lists are not common.

HoshinoTented: I don't know, ask @ice1000

HoshinoTented: I do think that ctors like `| suc {Nat}` look weird and useless, but we need to support `List {A}`

ice1000: I agree that `suc {Nat}` looks weird but it becomes less weird when the type becomes interesting (for example, the type of "less than" relation is isomorphic to Nat)

ice1000:
I believe that this reveals a serious problem in the design of `Pattern` ast. We may need to decouple the implicitness with the pattern itself

ice1000 (approved):
bors r+

base/src/test/java/org/aya/repr/ShapeMatcherTest.java
imkiva: Agreed!

imkiva:
> I believe that this reveals a serious problem in the design of `Pattern` ast. We may need to decouple the implicitness with the pattern itself

Recall the wired parsing code in `AyaGKProducer`, I can't agree more! We may need a `Arg<Pat>` (and also `Arg<Pattern>`)

imkiva:
bors merge

PR #641, opened 2022-11-08
Overhaul literal infrastructure

### LINKED ISSUE #598: Leftover of #589

imkiva:
Fix #598.

tl;dr: The unification now starts to play a role in solving literal types and disambiguation. RIP the old hacky and tricky code!

Details: When a literal may be represented by more than one datatype, we construct a hole for it and store it as the type in a `MetaLit`. Every time a _type-solved literal_ is required during unification, we eagerly check for the solution to the type and attempt to inline the `MetaLit` (like, when unfolding a global definition, normalizing a substituted term, etc.).

What's left: Because we need to know which universe list elements live in, which is only known when the underlying `DataCall` is provided, ambiguous list literals cannot yet be handled in this manner.


ice1000:
Conflicts

ice1000:
@imkiva conflict again

imkiva:
little task: cover the code below

image

ice1000:
Rest lgtm

base/src/main/java/org/aya/core/repr/AyaShape.java
```diff
@@ -36,15 +40,17 @@ record AyaIntLitShape() implements AyaShape {
   }
 
   record AyaListShape() implements AyaShape {
+    public static final @NotNull CodeShape.MomentId NIL = new CodeShape.MomentId();
+    public static final @NotNull CodeShape.MomentId CONS = new CodeShape.MomentId();
```
ice1000, line 44: Why not just use the all-in-one enum directly?

base/src/main/java/org/aya/core/repr/CodeShape.java
```diff
@@ -29,8 +38,9 @@ record StructShape(
   ) implements CodeShape {}
 
   record CtorShape(
+    @NotNull MomentId name,
     @NotNull ImmutableSeq<ParamShape> tele
-  ) implements CodeShape {}
+  ) implements CodeShape, Moment {}
```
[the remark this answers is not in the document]
ice1000: Ok I see I was stupid

base/src/main/java/org/aya/core/term/MetaLitTerm.java
```diff
@@ -0,0 +1,29 @@
...
+import org.aya.core.def.GenericDef;
+import org.aya.core.repr.AyaShape;
+import org.aya.core.repr.ShapeRecognition;
+import org.aya.util.error.SourcePos;
+import org.jetbrains.annotations.NotNull;
+
+public record MetaLitTerm(
+  @NotNull SourcePos sourcePos,
+  @NotNull Object repr,
+  @NotNull ImmutableSeq<Tuple2<GenericDef, ShapeRecognition>> candidates,
```
ice1000, line 17: ImmutableMap?

base/src/main/java/org/aya/core/term/MetaLitTerm.java
```diff
@@ -0,0 +1,29 @@
...
+import kala.collection.immutable.ImmutableSeq;
+import kala.control.Option;
+import kala.tuple.Tuple2;
+import org.aya.core.def.GenericDef;
+import org.aya.core.repr.AyaShape;
+import org.aya.core.repr.ShapeRecognition;
+import org.aya.util.error.SourcePos;
+import org.jetbrains.annotations.NotNull;
+
+public record MetaLitTerm(
```
ice1000, line 14: Is there a "metavar" for metalits?

base/src/main/java/org/aya/core/term/MetaLitTerm.java
HoshinoTented: I think it is possible that the same data matches different shapes.
Maybe `ImmutableMap<GenericDef, ImmutableSeq<ShapeRecognition>>` ?

ice1000: I doubt. This represents the "candidates" which I do not expect to have so many.

But also since it's just a list of candidates, I tend to agree with the current code

base/src/main/java/org/aya/core/term/MetaLitTerm.java
imkiva: `MetaLit` reuses the `MetaTerm` as its type, so the metavar is just ordinary meta (unlike `MetaPat`)

base/src/main/java/org/aya/core/repr/AyaShape.java
imkiva: Good point

HoshinoTented (approved):

base/src/main/java/org/aya/core/term/MetaLitTerm.java
ice1000: Cooooooooooooooool

PR #646, opened 2022-11-09
More Colorful Cli

HoshinoTented:
Question0: **Do we need to**/**How to** ser/deserialize Styles?
Question1: How does the user to modify the `RenderOptions`


ice1000:
> **Do we need to**/**How to** ser/deserialize Styles?

I don't believe so.

> How does the user to modify the `RenderOptions`

```
C:\Users\ice10\git-repos\aya-dev [luna ≡]> aya -i
Aya 0.23-SNAPSHOT (499bf9bb7fd2fa3546d3a9d0f69763967c810944)
> :print-toggle
Current pretty printing options:
InlineMetas: true
ShowImplicitArgs: false
ShowImplicitPats: true
ShowLambdaTypes: false
> :print-toggle ShowImplicitArgs
ShowImplicitArgs changed to true
> :unicode
Toggled Unicode to be disabled
> :unicode
Toggled Unicode to be enabled
>
```

ice1000:
Oh my bad. I misread your questions. For `RenderOptions`, I guess we can just ask users to use predefined styles. We don't need to be _that_ configurable.

ice1000:
We can choose to load styles from a json, and I believe that's not going to be very hard. For now, using predefined styles is already quite good

cli/src/main/java/org/aya/cli/repl/RenderOptions.java
```diff
@@ -0,0 +1,152 @@
...
+import java.lang.reflect.Type;
+import java.util.Map;
+
+public record RenderOptions(
+  @NotNull ColorScheme colorScheme,
+  @NotNull StyleFamily styleFamily
+) {
+  public final static class Deserializer implements JsonDeserializer<RenderOptions>, JsonSerializer<RenderOptions> {
+    public static class ColorSchemeJson {
+      public @Nullable java.util.Map<String, String> definedColors;
```
ice1000, line 26: Why full qualified name

cli/src/main/java/org/aya/cli/repl/RenderOptions.java
```diff
@@ -0,0 +1,152 @@
...
+
+    public @NotNull Reporter reporter;
+    public @NotNull RenderOptions fallback;
+
+    public Deserializer(@NotNull Reporter reporter, @NotNull RenderOptions fallback) {
+      this.reporter = reporter;
+      this.fallback = fallback;
+    }
+
+    public @NotNull ColorScheme colorSchemeFromJson(@NotNull ColorSchemeJson options) {
```
ice1000, line 42: Maybe use a data class directly? I don't think we'll be having custom colors

cli/src/main/java/org/aya/cli/repl/RenderOptions.java
HoshinoTented: Because I used both kala and java.util

cli/src/main/java/org/aya/cli/repl/RenderOptions.java
HoshinoTented: Did you mean the builtin color schemes only?

ice1000: I'm talking about the keys. As I understood, a color scheme is a set of key-value pairs where keys are some names and values are colours (or other attributes). So, we can simply make `class ColorScheme { Style fnName; Style dataName; ... }` and use Gson to deserialize it and then use it in the pretty printer

cli/src/main/java/org/aya/cli/repl/RenderOptions.java
ice1000: Might be a reason to migrate to jackson...

HoshinoTented: > Why `java.util`? Is it because it's used in Gson?

Yes.

ice1000:
How to modify the pretty printing options in the REPL?

HoshinoTented:
> How to modify the pretty printing options in the REPL?

I don't know. Maybe `:color propertyName "colorCode"` ?

ice1000:
> > How to modify the pretty printing options in the REPL?
> 
> I don't know. Maybe `:color propertyName "colorCode"` ?

Can you impwement this

HoshinoTented:
> > > How to modify the pretty printing options in the REPL?
> > 
> > 
> > I don't know. Maybe `:color propertyName "colorCode"` ?
> 
> Can you impwement this

I think I can.

ice1000:
Let's do it in separate PR

bors r+

PR #655, opened 2022-11-11
Fix some code

ice1000:
bors r+


HoshinoTented:
```diff
       // don't infer
-     if (localCtx == null) throw new Mismatch(true);
-     var bindSubst = new PatTraversal.MetaBind(this.subst, metaPat.ref().fakeBind().definition());
+     var bindSubst = new PatTraversal.MetaBind(subst, metaPat.ref().fakeBind().definition());
```

I don't think we can drop this case, consider this code:

```aya
open data Nat | O | S Nat
open data Vec (n : Nat)
| O => vnil
| S n => vcons (Vec n)

def inferMeta (n : Nat) : Nat
| O => O
| S n' => S n'

def justMatch {n : Nat} (xs : Vec (inferMeta n)) : Nat
| vnil => O
| vcons xs' => S O
```

When we are trying to tyck the Pat: `vcons xs'`, the corresponding signature is `Vec (inferMeta <O>)` instead of `Vec (inferMeta <n>)`, because the `Pat.Meta` `<n>` was inferred when normalizing `Vec (inferMeta <n>)`

HoshinoTented:
Sorry, my brain went into short circuit...

ice1000:
> the corresponding signature is Vec (inferMeta <O>)

Why O? Shouldn't it be S n?

ice1000:
> Sorry, my brain went into a short circuit...

What do you mean shortcut? Do you still agree with yourself?

HoshinoTented:
> > the corresponding signature is Vec (inferMeta )
> 
> Why O? Shouldn't it be S n?

There may be some problems which I haven't found out.

ice1000:
> > > the corresponding signature is Vec (inferMeta )
> > 
> > Why O? Shouldn't it be S n?
> 
> There may be some problems which I haven't found out.

IMO, Pat.Meta will be always inlined after PatTycker, so why bother handling them in PatClassifier?

HoshinoTented:
> > > > the corresponding signature is Vec (inferMeta )
> > > 
> > > 
> > > Why O? Shouldn't it be S n?
> > 
> > 
> > There may be some problems which I haven't found out.
> 
> IMO, Pat.Meta will be always inlined after PatTycker, so why bother handling them in PatClassifier or PatMatcher?

It is probably used during pat tyck, and it should not be inferred when used as an argument to another function.

ice1000:
I see now! Tyvm!

HoshinoTented:
> I see now! Tyvm!

But the code doesn't work as expected... I am trying to figure out the reason.

PR #658, opened 2022-11-14
Do Unifying Owner Telescope

HoshinoTented:


base/src/main/java/org/aya/core/def/CtorDef.java
```diff
@@ -40,17 +41,22 @@ public CtorDef(
...
-    var core = defVar.core;
-    if (core != null) return new DataDef.CtorTelescopes(core.ownerTele, core.selfTele);
-    var dataSignature = defVar.concrete.patternTele;
-    assert dataSignature != null;
-    var conSignature = defVar.concrete.signature;
-    assert conSignature != null;
-    return new DataDef.CtorTelescopes(dataSignature, conSignature.param());
+    var concrete = defVar.concrete;
+    if (concrete != null) {
+      // See StmtTycker.java#L222
```
ice1000, line 50: Do not use a line number, because it changes over time without notice. You can use a function name, or just copy the code.

ice1000: I mean copy the code in the comment

ice1000 (commented):
Rest lgtm

base/src/main/java/org/aya/core/serde/SerPat.java
```diff
@@ -43,14 +43,12 @@ record Bind(boolean explicit, @NotNull SerTerm.SimpVar var, @NotNull SerTerm ty)
...
-    @NotNull ImmutableSeq<SerTerm> ownerArgs,
     @NotNull ImmutableSeq<SerPat> params,
     @NotNull SerTerm.DataCall ty
   ) implements SerPat {
     @Override public @NotNull Pat de(SerTerm.@NotNull DeState state) {
       return new Pat.Ctor(explicit, state.resolve(name),
-        ownerArgs.map(arg -> arg.de(state)),
-        params.map(param -> param.de(state)),
+              params.map(param -> param.de(state)),
         ty.de(state));
```
imkiva, line 52: Format

base/src/main/java/org/aya/core/pat/Pat.java
```diff
@@ -154,35 +154,28 @@ record Tuple(
   }
 
   /**
-   * @param ownerArgs the arguments to the owner telescope of this Ctor.
+   *
    */
   record Ctor(
```
imkiva, line 159: Maybe delete the empty doc comment.

imkiva (approved):
nitpicks 😄

base/src/main/java/org/aya/core/serde/SerPat.java
HoshinoTented: ice1000's bad

base/src/main/java/org/aya/core/def/CtorDef.java
HoshinoTented: my bad

ice1000:
Are you sure this will work

HoshinoTented:
For code:

```aya
open data Nat | O | S Nat
open data Vec (n : Nat) (A : Type)
| O, A => vnil
| S n, A => vcons A (Vec n A)
      
def what => vnil {Nat}
def how => vcons {Nat} O vnil
```

Before fix:
image
image

After fix:
image
image

Also, `CtorTelescope` wasn't used anywhere except `ExprTycker`, so it is safe to modify it.

ice1000:
bors r= @ice1000 @imkiva 
Proof accepted!!

PR #673, opened 2022-11-17
graal: generate part of the config

ice1000:
@imkiva


base/build.gradle.kts
```diff
@@ -33,6 +33,11 @@ sourceSets.main {
 
 tasks.compileJava { dependsOn(generateVersion) }
 tasks.sourcesJar { dependsOn(generateVersion) }
+tasks.withType<org.aya.gradle.GenerateReflectionConfigTask>().configureEach {
+  extraDir = buildDir.resolve("classes/java/main/org/aya/core/serde")
+  classPrefixes = listOf("SerTerm$", "SerPat$", "SerDef$")
+  packageName = "org.aya.core.serde"
```
imkiva, line 39: What would happen if `listOf("SerTerm", "SerPat", "SerDef")` so everything under `org.aya.core.serde` can be generated?

imkiva (approved):

base/build.gradle.kts
ice1000: I don't know

ice1000: I don't understand the sentence

imkiva: I will take this. Don't worry 😄

imkiva:
Does this mean the generator won't work if we start a fresh build (like in CI)? Because we will not have `.class` files.

imkiva:
bors d=ice1000

PR #693, opened 2022-11-18
Fix my opinions

imkiva:
- `Doc.renderXXX` now uses the default configuration for each "output target".
- `RenderOptions.render()` now renders the `Doc` with configurable colors and styles.
- Whenever outputting to a terminal, use `AdaptiveCli` (by default) or `UnixTermStylist` (if either color or style is changed), which is guaranteed now by `RenderOptions.stylist()` and `RenderOptions.defaultStylist()`
- `AdaptiveCli` is now a standalone `Stylist` (instead of a `StyleFamily`)
- `Stylist` can be viewed as the frozen `Sig CS ** SF ** OutputTarget`, while `RenderOptions` reads user-configured `CS` and `SF` and create the `Stylist` according to "output target".

### DIFF
--- a/cli/src/main/java/org/aya/cli/render/RenderOptions.java
+++ b/cli/src/main/java/org/aya/cli/render/RenderOptions.java
@@ -1,9 +1,15 @@
 // Copyright (c) 2020-2022 Tesla (Yinsen) Zhang.
 // Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
-package org.aya.cli.repl.render;
+package org.aya.cli.render;
 
 import com.google.gson.JsonParseException;
-import org.aya.cli.repl.render.vscode.ColorTheme;
+import org.aya.cli.render.vscode.ColorTheme;
+import org.aya.pretty.backend.html.Html5Stylist;
+import org.aya.pretty.backend.latex.TeXStylist;
+import org.aya.pretty.backend.string.StringStylist;
+import org.aya.pretty.backend.string.style.AdaptiveCliStylist;
+import org.aya.pretty.backend.string.style.DebugStylist;
+import org.aya.pretty.backend.string.style.UnixTermStylist;
 import org.aya.pretty.printer.ColorScheme;
 import org.aya.pretty.printer.StyleFamily;
 import org.aya.pretty.style.AyaColorScheme;
@@ -16,7 +22,15 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Objects;
 
+/** Multi-target {@link org.aya.pretty.printer.Stylist}, usually created from json files. */
 public class RenderOptions {
+  public enum OutputTarget {
+    Terminal,
+    LaTeX,
+    HTML,
+    Other,
+  }
+
   public enum ColorSchemeName {
     Emacs,
     IntelliJ,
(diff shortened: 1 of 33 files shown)


cli/src/main/java/org/aya/cli/render/RenderOptions.java
```diff
@@ -0,0 +1,131 @@
...
+  public void checkInitialize() {
+    if (colorScheme == null) colorScheme = DEFAULT_COLOR_SCHEME;
+    if (styleFamily == null) styleFamily = DEFAULT_STYLE_FAMILY;
+  }
+
+  public boolean isDefault() {
+    return colorScheme.equals(DEFAULT_COLOR_SCHEME) && styleFamily.equals(DEFAULT_STYLE_FAMILY);
+  }
+
+  public @NotNull StringStylist defaultStylist(@NotNull OutputTarget output) {
```
HoshinoTented, line 70: Maybe `static`?

HoshinoTented (approved):
LGTM!!

cli/src/main/java/org/aya/cli/repl/ReplConfig.java
```diff
@@ -36,11 +35,10 @@ public class ReplConfig implements AutoCloseable {
    */
   public @UnknownNullability RenderOptions renderOptions = new RenderOptions();
   public transient @NotNull StringStylist stylist;
```
HoshinoTented, line 37: Maybe `private` or remove the useless `getStylist()`

imkiva: Actually we do not need to store the stylist here. But now it behaves a little bit different: it parses the vsc file every time we render a Doc. I am not sure if it a big performance issue....

imkiva: We may use a LazyValue with expiration check to store the stylist if so.

HoshinoTented: > Actually we do not need to store the stylist here. But now it behaves a little bit different: it parses the vsc file every time we render a Doc. I am not sure if it a big performance issue....

I think the `stylist` should be memorized.

imkiva: Then it should be stored in RenderOptions itself.

I am fixing it tomorrow.

cli/src/main/java/org/aya/cli/single/SingleFileCompiler.java
```diff
@@ -95,16 +87,15 @@ private void distill(
...
-        new DocHtmlPrinter.Config(scm, AyaStyleFamily.DEFAULT, b)));
-      case latex -> doWrite(doc, distillDir, fileName, ".tex", (d, $) -> d.render(new DocTeXPrinter(),
-        new DocTeXPrinter.Config(scm, AyaStyleFamily.DEFAULT)));
-      case plain -> doWrite(doc, distillDir, fileName, ".txt", (d, $) -> d.debugRender());
-      case unix -> doWrite(doc, distillDir, fileName, ".txt", (d, $) -> d.renderToString(
-        StringPrinterConfig.unixTerminal(scm, AyaStyleFamily.DEFAULT, PrinterConfig.INFINITE_SIZE, true)));
-    }
+    var renderOptions = flags.renderOptions();
+    var out = switch (flags.distillFormat()) {
+      case html -> Tuple.of(".html",RenderOptions.OutputTarget.HTML);
```
ice1000, line 92: Fmt

cli/src/main/java/org/aya/cli/single/SingleFileCompiler.java
```diff
@@ -95,16 +87,15 @@ private void distill(
...
+      case latex -> Tuple.of(".tex",RenderOptions.OutputTarget.LaTeX);
+      case plain -> Tuple.of(".txt",RenderOptions.OutputTarget.Plain);
+      case unix -> Tuple.of(".txt",RenderOptions.OutputTarget.Terminal);
```
ice1000, line 95: Can these file extensions be fields of OutputTarget

ice1000 (approved):

cli/src/main/java/org/aya/cli/repl/ReplConfig.java
ice1000: You are fixing it in the next PR

PR #699, opened 2022-11-19
Let Expression

HoshinoTented:


HoshinoTented:
The new syntax makes Grammar-Kit unhappy, how to improve it?

HoshinoTented:
Thanks @imkiva !!

HoshinoTented:
The let-expression isn't a syntax sugar now, should it?

HoshinoTented:
FINAL QUESTION:
`=>` or `:=` or others

@imkiva @ice1000 @lunalunaa

HoshinoTented:
One problem is that we can use `let` in type (I think so does `match`). Is it allowed?

HoshinoTented:
> One problem is that we can use `let` in type (I think so does `match`). Is it allowed?

Also, do we need to allow declaring a function by `let`, like `let f a := g a a in f x`?
For the non-recursion function, it is trivial: `let f := \ a => g a a in f x`, but the problem is the recursion function. Do we have something like `fix` in Coq?

ice1000 (commented):
Can I propose another syntax? What do you think about let | a b := c d | e f := g h which is the same as partial

HoshinoTented:
> Can I propose another syntax? What do you think about let | a b := c d | e f := g h which is the same as partial

So we should able to define functions by the `let` expression?

ice1000:
> > Can I propose another syntax? What do you think about let | a b := c d | e f := g h which is the same as partial
> 
> So we should able to define functions by the `let` expression?

I expected function definitions, otherwise `let` is too limited. You can always use lambdas though.

HoshinoTented:
> > > Can I propose another syntax? What do you think about let | a b := c d | e f := g h which is the same as partial
> > 
> > 
> > So we should able to define functions by the `let` expression?
> 
> I expected function definitions, otherwise `let` is too limited. You can always use lambdas though.

Then the syntax will look like:

```aya
let
  | f (x : X) : F := g x
in f y
```

which is equivalent to:

```aya
(\ (f : X -> F) => f y) (\ (x : X) => g x)
```

ice1000:
Where did the `: F` go?

HoshinoTented:
> Where did the `: F` go?

`f : X -> F`

ice1000:
> > One problem is that we can use `let` in type (I think so does `match`). Is it allowed?
> 
> Also, do we need to allow declaring a function by `let`, like `let f a := g a a in f x`? For the non-recursion function, it is trivial: `let f := \ a => g a a in f x`, but the problem is the recursion function. Do we have something like `fix` in Coq?

No. I expect no recursive functions over there.

Note that we may in the future extend this functionality a little bit. Say, we can define a `let x := y` where `x` will _not_ reduce in the function body.

ice1000:
Fanzheng Arend has implemented that. I conjecture that they did it for reasons.

HoshinoTented:
TODO: update distiller

ice1000:
Merge main?

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -848,17 +849,43 @@ private Result unifyTyMaybeInsert(@NotNull Term upper, @NotNull Result result, E
     return checker.apply(term);
   }
 
-  /// region Helper
+  /**
+   * @param letBinds not empty
```
imkiva, line 853: But you actually handled the empty case, right?
I suggest remove this comment.

base/src/main/java/org/aya/tyck/ExprTycker.java
```diff
@@ -848,17 +849,43 @@ private Result unifyTyMaybeInsert(@NotNull Term upper, @NotNull Result result, E
...
 
-  public <R> R subscoped(@NotNull Supplier<R> action) {
     var parent = this.localCtx;
     this.localCtx = parent.deriveMap();
-    var result = action.get();
+    this.localCtx.put(param);
+
+    var result = desugarLet(remain, leaf);
+
     this.localCtx = parent;
```
imkiva, line 879: `.subscoped` ?

base/src/main/java/org/aya/core/term/Term.java
```diff
@@ -40,7 +40,7 @@
  * @author ice1000
  */
 public sealed interface Term extends AyaDocile, Restr.TermLike<Term>
-  permits Callable, CoeTerm, Elimination, ErasedTerm, FormulaTerm, HCompTerm, IntervalTerm, MatchTerm, MetaPatTerm, MetaLitTerm, PartialTerm, PiTerm, RefTerm, RefTerm.Field, SigmaTerm, StableWHNF {
+        permits Callable, CoeTerm, Elimination, ErasedTerm, FormulaTerm, HCompTerm, IntervalTerm, MatchTerm, MetaLitTerm, MetaPatTerm, PartialTerm, PiTerm, RefTerm, RefTerm.Field, SigmaTerm, StableWHNF {
```
imkiva, line 43: formats

base/src/test/resources/success/src/LetExpr.aya
```diff
@@ -0,0 +1,36 @@
...
+  let
+  | n1 := suc n
+  | n2 := suc n1
+  in n2
+
+example def normalize : slow_plus2 7 = 9 => idp
+
+// TODO: should we allow this?
+
+def letInType : (let a := 1 in a = a) => idp
```
imkiva, line 14: Yes. They seems quite innocent.

cli/src/main/java/org/aya/cli/parse/AyaGKProducer.java
```diff
@@ -792,6 +800,17 @@ private Arg<Pattern> unitPattern(@NotNull GenericNode<?> node) {
     return Expr.Array.newList(entireSourcePos, exprs);
   }
 
+  public @NotNull Expr.Let.Bind letBind(@NotNull GenericNode<?> node) {
+    var pos = sourcePosOf(node);
+    var bind = weakId(node.child(WEAK_ID));
+    // make IDEA happy
+    var teles = lambdaTelescope((SeqView<GenericNode<?>>) node.childrenOfType(LAMBDA_TELE));
```
imkiva, line 807: `node.childrenOfType(...).map(x -> x)`

base/src/main/java/org/aya/tyck/ExprTycker.java
HoshinoTented: My bad

cli/src/main/java/org/aya/cli/parse/AyaGKProducer.java
HoshinoTented: My stupid brain rejected this code before. 😭

base/src/main/java/org/aya/core/term/Term.java
HoshinoTented: 😭

cli/src/main/java/org/aya/cli/parse/AyaGKProducer.java
HoshinoTented: My stupid brain said the `.map(x -> x)` is inefficient.

ice1000: It is but I don't think it matters that much

ice1000 (approved):

cli/src/main/java/org/aya/cli/parse/AyaGKProducer.java
HoshinoTented: > It is but I don't think it matters that much

Yes, my bad habit.

imkiva: I think `.map` is better as it only causes compile errors, while type cast may throw runtime exceptions.

HoshinoTented: > I think `.map` is better as it only causes compile errors, while type cast may throw runtime exceptions.

YES! I got it.

HoshinoTented:
bors d=imkiva

imkiva (dismissed):
bors d=HoshinoTented

imkiva:
bors merge

ice1000:
> Merge conflict.

Why

HoshinoTented:
> > Merge conflict.
> 
> Why

Parser

ice1000:
bors merge

PR #700, opened 2022-11-19
Kiva fixing something today

### LINKED ISSUE #694: Kiva fixing something tomorrow
Then it should be stored in RenderOptions itself.

I am fixing it tomorrow.

_Originally posted by @imkiva in https://github.com/aya-prover/aya-dev/pull/693#discussion_r1026696499_

imkiva:
fix #694

- Memorize `stylist` for each output target
- Strongly typed argument for `:color` and `:style`, as well as completions!

### DIFF
--- a/cli/src/main/java/org/aya/cli/render/RenderOptions.java
+++ b/cli/src/main/java/org/aya/cli/render/RenderOptions.java
@@ -3,8 +3,8 @@
 package org.aya.cli.render;
 
 import com.google.gson.JsonParseException;
-import kala.collection.Seq;
 import kala.collection.mutable.MutableHashMap;
+import kala.control.Either;
 import kala.control.Try;
 import org.aya.cli.render.vscode.ColorTheme;
 import org.aya.pretty.backend.html.DocHtmlPrinter;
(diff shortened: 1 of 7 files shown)


imkiva:
The 700th!

image

cli/src/main/java/org/aya/cli/render/RenderOptions.java
```diff
@@ -83,13 +119,15 @@ public boolean isDefault() {
...
-      case HTML -> new Html5Stylist(buildColorScheme(), buildStyleFamily());
-      case Plain -> new DebugStylist(buildColorScheme(), buildStyleFamily());
-    };
+    return stylistCaches.getOrPut(output, () -> Try.of(() -> {
+      if (isDefault()) return defaultStylist(output);
+      return switch (output) {
+        case Terminal -> new UnixTermStylist(buildColorScheme(), buildStyleFamily());
+        case LaTeX -> new TeXStylist(buildColorScheme(), buildStyleFamily());
+        case HTML -> new Html5Stylist(buildColorScheme(), buildStyleFamily());
+        case Plain -> new DebugStylist(buildColorScheme(), buildStyleFamily());
```
HoshinoTented, line 128: It seems to build the same color scheme again when we require the stylist of another `OutputTarget`.

imkiva: Tql

cli/src/main/java/org/aya/cli/render/RenderOptions.java
```diff
@@ -73,6 +77,38 @@ public boolean isDefault() {
...
+
+  public void updateStyleFamily(@NotNull Either<StyleFamilyName, Path> nameOrPath) throws IllegalArgumentException {
+    if (nameOrPath.isRight()) throw new IllegalArgumentException("We don't support custom style family yet :)");
+    this.styleFamily = nameOrPath.getLeftValue();
+    invalidate();
+  }
+
+  public void invalidate() {
+    colorSchemeCache = null;
+    styleFamilyCache = null;
```
HoshinoTented, line 109: Maybe we should invalidate them separately? However, this is okay for now because we can only change the color scheme.

HoshinoTented (approved):
LGTM!!

cli/src/main/java/org/aya/cli/render/RenderOptions.java
imkiva: I don't think it is a good idea, as it makes the status of render options much more complicated.

imkiva: And we don't care about the extra performance overhead. It must not be the bottleneck.

imkiva:
bors merge

cli/src/main/java/org/aya/cli/render/RenderOptions.java
HoshinoTented: You persuaded me.

PR #702, opened 2022-11-19
New `group` library config and polish some messages

imkiva:
as title

### DIFF
--- a/base/src/test/resources/success/aya.json
+++ b/base/src/test/resources/success/aya.json
@@ -1,5 +1,6 @@
 {
   "ayaVersion": "0.23",
+  "group": "org.aya-prover",
   "name": "tests",
   "version": "0.1.0",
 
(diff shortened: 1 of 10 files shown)


ice1000:
But what's the point of `group`?

imkiva:
> But what's the point of `group`?

1. Putting all packages in the global namespace results in name clash, or name squatting, see crates.io. 

2. If we are going to have a centralized package hosting server, we must have scoped packages. If we also support decentralized one, take GitHub for example, the GitHub user name is itself a scope.

3. (not that important, not the only approach) If we are compling aya to JVM (or GraalVM), we should have the ability to interoperate between java and aya. In this case, I will compile `Mod::Hello` in the package "org.aya-prover" to a class "org.ayaprover.mod.Hello" for java's convenience.

PR #714, opened 2022-11-21
Overhaul `loadFile` in REPL

imkiva:
fix: repl cannot use shapes and operator bindings after `:l file.aya`

reported in a QQ group. /cc @ShreckYe


imkiva:
bors merge

PR #715, opened 2022-11-21
Fix SourcePos of Let

HoshinoTented:
image

### DIFF
--- a/base/src/main/java/org/aya/concrete/Expr.java
+++ b/base/src/main/java/org/aya/concrete/Expr.java
@@ -712,11 +712,31 @@ public sealed interface Expr extends AyaDocile, SourceNode, Restr.TermLike<Expr>
    *   <li>{@link Let.Bind#definedAs} = g</li>
    *   <li>{@link Let#body} = expr</li>
    * </ul>
+   *
+   * @param sourcePos to the source pos of body (source code level), consider this code
+   *                  <pre>
+   *                  let a := b in
+   *                  let c := d in
+   *                  e
+   *                  </pre>
+   *                  The source pos of the
+   *                  `let a := b in let c := d in e`
+   *                  and
+   *                  `let c := d in e`
+   *                  are both the source pos of e.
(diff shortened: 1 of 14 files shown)


base/src/main/java/org/aya/concrete/Expr.java
```diff
@@ -708,15 +708,36 @@ public static Expr.Array newGenerator(
...
   record Let(
+    @NotNull SourcePos sourcePos,
     @NotNull Let.Bind bind,
     @NotNull Expr body
   ) implements Expr {
+    public Let {
+      if (body instanceof Let let) {
+        sourcePos = let.sourcePos();
+      }
+    }
```
imkiva, line 739: The syntax location should not be the same as the error reporting location. The correct `sourcePos` of a `Let` should always be `sourcePosOf(GenericNode)` in the producer. Or some location-based features may not work freely for `Let`

HoshinoTented: So what do we expect? The whole let-expression or the body?

ice1000: 0. Let should have whole SourcePos
1. LetBind can have another SourcePos and we use that for error report

HoshinoTented: So: For all `Let`s in the same group, they share the same source pos -- the whole source pos

HoshinoTented: I think it is good to have a way to obtain the source pos of an expr that determines the type of the whole expr (like the last `DoBind` of a do-notation, the body of `Let`)

imkiva: > So: For all `Let`s in the same group, they share the same source pos -- the whole source pos

No

imkiva: > I think it is good to have a way to obtain the source pos of an expr that determines the type of the whole expr (like the last `DoBind` of a do-notation, the body of `Let`)

Yes

PR #716, opened 2022-11-22
Fix Highlighter

HoshinoTented:
There are some problems with `Highlighter`

* `Expr.Param#ref`s didn't highlight (so no definitions)
* For the parameters that share the same type, like `(a b : A)`, the type will be highlighted twice, we can perform `.distinct` after `.sort` to solve this problem

### DIFF
--- a/cli/src/main/java/org/aya/cli/literate/SyntaxHighlight.java
+++ b/cli/src/main/java/org/aya/cli/literate/SyntaxHighlight.java
@@ -96,7 +96,7 @@ public class SyntaxHighlight implements StmtFolder<MutableList<HighlightInfo>> {
         if (decl instanceof Decl.Telescopic teleDecl) {
           teleDecl.telescope().view()
             .map(Expr.Param::ref)
-            .filter(x -> !x.isGenerated())
+            .filterNot(LocalVar::isGenerated)
             .forEach(def -> add(acc, linkDef(def.definition(), def)));
         }
 
--- a/cli/src/main/java/org/aya/cli/literate/SyntaxHighlight.java
+++ b/cli/src/main/java/org/aya/cli/literate/SyntaxHighlight.java
@@ -93,9 +93,8 @@ public class SyntaxHighlight implements StmtFolder<MutableList<HighlightInfo>> {
         if (decl instanceof Decl.Telescopic teleDecl) {
           teleDecl.telescope().view()
             .map(Expr.Param::ref)
-            .filter(x -> x != LocalVar.IGNORED && !x.isGenerated())
-            .forEach(def ->
-              add(acc, linkDef(def.definition(), def)));
+            .filter(x -> !x.isGenerated())
+            .forEach(def -> add(acc, linkDef(def.definition(), def)));
         }
 
         yield add(acc, linkDef(decl.sourcePos(), decl.ref()));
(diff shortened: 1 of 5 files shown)


cli/src/main/java/org/aya/cli/literate/SyntaxHighlight.java
```diff
@@ -89,7 +89,17 @@ public class SyntaxHighlight implements StmtFolder<MutableList<HighlightInfo>> {
       case Command.Open o when o.fromSugar() -> acc;  // handled in `case Decl` or `case Command.Import`
       case Command.Open o -> add(acc, linkModuleRef(o.path()));
       case ClassDecl decl -> add(acc, linkDef(decl.sourcePos, decl.ref()));
-      case Decl decl -> add(acc, linkDef(decl.sourcePos(), decl.ref()));
+      case Decl decl -> {
+        if (decl instanceof Decl.Telescopic teleDecl) {
+          teleDecl.telescope().view()
+            .map(Expr.Param::ref)
+            .filter(x -> x != LocalVar.IGNORED && !x.isGenerated())
```
imkiva, line 96: Construct `LocalVar.IGNORED` with `isGenerated = true`

cli/src/main/java/org/aya/cli/literate/SyntaxHighlight.java
```diff
@@ -89,7 +89,17 @@ public class SyntaxHighlight implements StmtFolder<MutableList<HighlightInfo>> {
...
+            .forEach(def ->
+              add(acc, linkDef(def.definition(), def)));
```
imkiva, line 98: Just put them on the same line.

imkiva:
I kinda think we should not `.sorted().distinct()` every time we call highlighter, instead, let callers sort and distinct them. Because there are places we don't care about the order and duplication (like LSP).

HoshinoTented:
> I kinda think we should not `.sorted().distinct()` every time we call highlighter, instead, let callers sort and distinct them. Because there are places we don't care about the order and duplication (like LSP).

I agree.

RomanticD:
lgtm

tsao-chi:
lgtm
p.s. 不小心用同学的gh号评论了

cli/src/main/java/org/aya/cli/literate/SyntaxHighlight.java
```diff
@@ -89,7 +92,16 @@ public class SyntaxHighlight implements StmtFolder<MutableList<HighlightInfo>> {
       case Command.Open o when o.fromSugar() -> acc;  // handled in `case Decl` or `case Command.Import`
       case Command.Open o -> add(acc, linkModuleRef(o.path()));
       case ClassDecl decl -> add(acc, linkDef(decl.sourcePos, decl.ref()));
-      case Decl decl -> add(acc, linkDef(decl.sourcePos(), decl.ref()));
+      case Decl decl -> {
+        if (decl instanceof Decl.Telescopic teleDecl) {
+          teleDecl.telescope().view()
+            .map(Expr.Param::ref)
+            .filter(x -> !x.isGenerated())
```
ice1000, line 99:
```suggestion
            .filterNot(x -> x.isGenerated())
```

Is this available

ice1000:
bors r+
KISS

PR #720, opened 2022-11-23
cli: remove a duplicate checkDeserialization() call

tsao-chi:
It is already in `asConfig(@NotNull Path libraryRoot, @NotNull Function<String, Path> buildRootGen)`


ice1000:
Please add milestone and labels new time

PR #730, opened 2022-11-25
Improve pushing telescopes

imkiva:
We generate a bind pattern when the right-hand side is not a lambda. So the following code
```aya
def overlap infix + : Nat -> Nat -> Nat
  | 0     => id
  | suc a => \b => suc (a + b)
```
is desugared to

```aya
def overlap infix + Nat Nat : Nat
  | 0,     x => id x
  | suc a, b => suc (a + b)
```

which works with overlapping patterns.


ice1000 (changes requested):
image
No need to do this! We never need to push any patterns upon insufficiency, because absurdity will take care of it.

imkiva:
> image
> 
> No need to do this! We never need to push any patterns upon insufficiency, because absurdity will take care of it.

1. This code is taking care of absurdity. Just check the test file and you will know why this case is possible.

2. I said I am going to change the implementation in DM (checkRhs and insert missing patterns and then classify). The code is just outdated.

imkiva:
@ice1000 I decided to close this because classification should be done before checking RHS for pattern refinement.

PR #734, opened 2022-11-26
Remove ulift

ice1000:
/cc @tsao-chi


base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
```diff
@@ -66,11 +68,11 @@ public sealed abstract class TeleDecl extends CommonDecl implements Decl.Telesco
     this.telescope = telescope;
   }
 
-  @Override public @Nullable Def.Signature signature() {
-    return signature;
+  @Override public Def.@Nullable Signature<RetTy> signature() {
+    return null;
```
tsao-chi, line 72: Why null

ice1000: Nice catch!

PR #763, opened 2022-12-06
Fix a serious termination check bug

imkiva:
Aya rejects

```aya
def just-pred (n : Nat) : Nat => pred n

def pred Nat : Nat
  | zero => zero
  | suc n => just-pred n
```


imkiva:
But this code is still rejected after the fix commit

```aya
def just-pred => pred

def pred Nat : Nat
  | zero => zero
  | suc n => just-pred n
```

ice1000 (approved):
merge?

imkiva:
bors merge

PR #774, opened 2022-12-07
Cleanup, more about ordinals

ice1000:
Fixed some bugs on meta pats, added some lemmas about ordinals


ice1000:
Improved handling of missing bind patterns!!

bors r+

PR #777, opened 2022-12-07
Update Folder in the concrete layer.

wsx-ucb:
This allows finer grain control on the handling of variables, which we can now distinguish between variable declaration and reference.
The first commit solves an immediate design problem noticed by @HoshinoTented.
I've then noticed some duplicated logic that can be better handled by the new API in the LSP highlighter and resolver.
If no one is working on that part of the code currently, I can clean that up in a follow up commit.


ice1000:
> I've then noticed some duplicated logic that can be better handled by the new API in the LSP highlighter and resolver.

Can you do in separate PR? :wink: I would like to merge this early

PR #794, opened 2022-12-10
Extract more to ide module

imkiva:
@ice1000


ice1000:
Can you also solve #795?

ice1000 (dismissed):
Looks good

ide/src/main/java/org/aya/ide/action/Folding.java
```diff
@@ -0,0 +1,42 @@
...
+    var program = source.program().get();
+    if (program != null) program.forEach(folder);
+    return folder.foldingRanges.toImmutableSeq();
+  }
+
+  @Override public void accept(@NotNull Stmt stmt) {
+    switch (stmt) {
+      case Decl maybe -> Resolver.withChildren(maybe)
+        .filter(dv -> dv.concrete != null)
+        .map(dv -> dv.concrete.entireSourcePos())
```
ice1000, line 28:
```suggestion
        .mapNotNull(dv -> dv.concrete)
        .map(dv -> dv.entireSourcePos())
```

And then use method ref

imkiva:
> Can you also solve #795?

We will make `Reporter` and `Problem` generic types, is this intended?

imkiva:
I would make `DistillerOptions` a generic type 😉

ice1000:
> > Can you also solve #795?
> 
> We will make `Reporter` and `Problem` generic types, is this intended?

I think you can. Does that change a lot of existing code? 😟

ice1000:
Well, I just do not think it's a good idea to have Aya-specific stuffs in the `tools` module. It should be just tools.

imkiva:
bors d=ice1000

ice1000 (approved):
bors merge

PR #797, opened 2022-12-10
Use html "class" attribute to set code style

imkiva:
It is now customizable by frontends!


ice1000 (approved):
bors merge

ice1000:
Do you think HTML constants should go to some resource files?

imkiva:
> Do you think HTML constants should go to some resource files?

I am not sure. Some constants refer to each other, so it may be hard (or make the usage code longer) to extract.

ice1000:
> > Do you think HTML constants should go to some resource files?
> 
> I am not sure. Some constants refer to each other, so it may be hard (or make the usage code longer) to extract.

We can extract the basic ones

PR #799, opened 2022-12-10
fix(pattern): add check for absurd patterns

### LINKED ISSUE #775: We didn't handle absurd clauses without absurd patterns
```
data False
open data Truth | rua
def f Truth : False
| rua
def inconsistency : False => f rua
```

SchrodingerZhu:
- Gate away the reachable match arms with empty body from passing type-checking:
- Check: for those patterns used in elimination (rather than data decl), if Pattern.Absurd () does not appear on lhs, then we require a body

close #775

### DIFF
--- a/base/src/main/java/org/aya/tyck/pat/PatternProblem.java
+++ b/base/src/main/java/org/aya/tyck/pat/PatternProblem.java
@@ -143,4 +143,20 @@ public sealed interface PatternProblem extends Problem {
         Doc.par(1, type.toDoc(options)));
     }
   }
+
+  record InvalidEmptyBody(
+    @NotNull Pattern.Clause match
+  ) implements PatternProblem {
+    @Override @NotNull public Pattern pattern() {
+      return match.patterns.first().term();
+    }
+
+    @Override @NotNull public SourcePos sourcePos() {
+      return match.sourcePos;
+    }
+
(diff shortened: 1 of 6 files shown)


ice1000 (approved):
bors r+

ice1000 (commented):
bors r-

base/src/main/java/org/aya/tyck/pat/PatternProblem.java
```diff
@@ -143,4 +143,20 @@ record BadLitPattern(
         Doc.par(1, type.toDoc(options)));
     }
   }
+
+  record InvalidEmptyBody(
+    @NotNull Pattern.Clause match
+  ) implements PatternProblem {
+    @Override @NotNull public Pattern pattern() {
+      return match.patterns.first().term();
+    }
```
ice1000, line 152:
```suggestion
  ) implements Problem {
```

ice1000 (changes requested):
bors r-

base/src/main/java/org/aya/tyck/pat/PatternProblem.java
SchrodingerZhu: resolved

ice1000 (approved):
bors r+

PR #801, opened 2022-12-11
Improve pattern var folder

ice1000:
/cc @imkiva @wsx-ucb Is this correct?

### DIFF
--- a/base/src/main/java/org/aya/concrete/visitor/PatternFolder.java
+++ b/base/src/main/java/org/aya/concrete/visitor/PatternFolder.java
@@ -10,6 +10,7 @@ import org.aya.core.term.PiTerm;
 import org.aya.core.term.Term;
 import org.aya.ref.AnyVar;
 import org.aya.ref.DefVar;
+import org.aya.util.Arg;
 import org.aya.util.error.SourcePos;
 import org.jetbrains.annotations.MustBeInvokedByOverriders;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


base/src/main/java/org/aya/concrete/visitor/PatternFolder.java
```diff
@@ -30,15 +31,26 @@
...
+  default @NotNull R fold(@NotNull R acc, @NotNull Arg<Pattern> pat) {
+    return fold(acc, pat.term());
+  }
+
   @MustBeInvokedByOverriders
   default @NotNull R fold(@NotNull R acc, @NotNull Pattern pat) {
     return switch (pat) {
+      case Pattern.List list -> list.elements().foldLeft(acc, this::fold);
+      case Pattern.Tuple tuple -> tuple.patterns().foldLeft(acc, this::fold);
+      case Pattern.BinOpSeq seq -> seq.seq().foldLeft(acc, this::fold);
```
imkiva, line 43: Traverse nested patterns are done in 

https://github.com/aya-prover/aya-dev/blob/723d709514b067719746c0a98f92c4dde1feb3fc/base/src/main/java/org/aya/concrete/visitor/StmtFolder.java#L61-L80

So it is not needed here.

imkiva (commented):
This is not correct I suppose.

base/src/main/java/org/aya/concrete/visitor/PatternFolder.java
imkiva: Same for `Expr` and `Stmt`

ice1000:
bors close

ice1000:
Oh I can close it directly LMAO

PR #803, opened 2022-12-11
Just use markdown list for plain text backend

imkiva:
Markdown is already easy to read in plain text format.


ice1000 (commented):
Looks cool

ice1000 (dismissed):

imkiva:
bors merge

PR #806, opened 2022-12-11
Partially fix page width problem

### LINKED ISSUE #805: Doc page width problem
Try the following case:

```java
var d = Doc.english("You are right, but \"Genshin Impact\" is a brand new open world adventure game by MHY.");
System.out.println(d.renderToString(16, false));
```

Doc will be rendered as

```
You are right, but
"Genshin Impact"
is a brand new open
world adventure game
by MHY
```

There are lines exceeding the width of 16.

imkiva:
fix #805 

Just fixed the case with `Doc.english`, you can still construct a counterexample with `Doc.sep(Doc.plain(), Doc.plain(), ...)`


imkiva:
Codecov is really buggy....

Issue #821, opened 2022-12-14
Another native image bug


tsao-chi:
```
./cli/build/native/nativeCompile/aya -i
Aya 0.26-SNAPSHOT (29065eaf5572455d2cdd68b09df2f8759e61102d)
> :load base/src/test/resources/success
Compiling common
  [Info] Resolving source file dependency
    Done in 231ms
  [Tyck] Relation::Formula (Relation/Formula.aya)
com.oracle.svm.core.jdk.UnsupportedFeatureError: SerializationConstructorAccessor class not found for declaringClass: kala.function.Hasher$Optimized (targetConstructorClass: java.lang.Object). Usually adding kala.function.Hasher$Optimized to serialization-config.json fixes the problem.
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.VMError.unsupportedFeature(VMError.java:89)
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.reflect.serialize.SerializationSupport.getSerializationConstructorAccessor(SerializationSupport.java:143)
        at java.base@17.0.5/jdk.internal.reflect.MethodAccessorGenerator.generateSerializationConstructor(MethodAccessorGenerator.java:48)
        at java.base@17.0.5/jdk.internal.reflect.ReflectionFactory.generateConstructor(ReflectionFactory.java:463)
        at java.base@17.0.5/jdk.internal.reflect.ReflectionFactory.newConstructorForSerialization(ReflectionFactory.java:455)
        at java.base@17.0.5/java.io.ObjectStreamClass.getSerializableConstructor(ObjectStreamClass.java:1443)
        at java.base@17.0.5/java.io.ObjectStreamClass$2.run(ObjectStreamClass.java:412)
        at java.base@17.0.5/java.io.ObjectStreamClass$2.run(ObjectStreamClass.java:384)
        at java.base@17.0.5/java.security.AccessController.executePrivileged(AccessController.java:168)
        at java.base@17.0.5/java.security.AccessController.doPrivileged(AccessController.java:318)
        at java.base@17.0.5/java.io.ObjectStreamClass.<init>(ObjectStreamClass.java:384)
        at java.base@17.0.5/java.io.ObjectStreamClass$Caches$1.computeValue(ObjectStreamClass.java:110)
        at java.base@17.0.5/java.io.ObjectStreamClass$Caches$1.computeValue(ObjectStreamClass.java:107)
[...]

PR #822, opened 2022-12-15
nativeimg: update reflect-config

tsao-chi:
fix https://github.com/aya-prover/aya-dev/issues/821
bors r+


ice1000:
Cool!

What about testing it in CI?

tsao-chi:
ok

Issue #828, opened 2022-12-17
Rename distill to pretty


ice1000:
... everywhere because the initial meaning of distillation in Jon Sterling's project (https://github.com/RedPRL/redtt/issues/164) is to convert core to concrete. We didn't do that :(

PR #833, opened 2022-12-17
refactor: distill to pretty

tsao-chi:
fix https://github.com/aya-prover/aya-dev/issues/828


ice1000:
I believe that you need a different package name. The `pretty` module already occupied the `org.aya.pretty` package, or did it?

ice1000:
Yeah I think it did.

ice1000:
What about this: `org.aya.prettier`?

If you have a better idea, feel free to use it 😉

tsao-chi:
> What about this: `org.aya.prettier`?


ok

base/src/main/java/org/aya/concrete/desugar/BinExprParser.java
```diff
@@ -6,7 +6,7 @@
 import org.aya.concrete.Expr;
 import org.aya.concrete.error.OperatorError;
 import org.aya.generic.Constants;
-import org.aya.pretty.doc.Doc;
+import org.aya.prettier.doc.Doc;
 import org.aya.ref.DefVar;
 import org.aya.resolve.ResolveInfo;
```
imkiva, line 11: Please don't change `org.aya.pretty` in the pretty module.

imkiva (changes requested):
Still possible package name clashes: "org.aya.prettier" exist in both "pretty" and "base"

base/src/main/java/org/aya/concrete/desugar/BinExprParser.java
tsao-chi: oh

imkiva (approved):
bors d=tsao-chi

ice1000:
Thank you!!

PR #836, opened 2022-12-18
Fix Using

HoshinoTented:


ice1000:
:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

base/src/main/java/org/aya/resolve/context/ModuleExport.java
```diff
@@ -0,0 +1,176 @@
...
+   * @return false if failed
+   */
+  public boolean export(@NotNull String name, @NotNull AnyVar ref) {
+    var exists = putIfAbsent(name, ref);
+    return exists.isEmpty();
+  }
+
+  // TODO: delete this garbage
+  @Deprecated
+  public @NotNull Result filter(@NotNull ImmutableSeq<String> names, boolean isUse, @NotNull SourcePos pos) {
```
HoshinoTented, line 38: I hope the `Name`s in UseHide are `WithPos`.

ice1000: These names are also serialized, so they do not always have `SourcePos`. Is there another way around?

ice1000: But done.

ice1000 (approved):
bors r+

ice1000:
Thank you!

HoshinoTented:
> Thank you!

I think we need to wait for imkiva's review.

ice1000:
> > Thank you!
> 
> I think we need to wait for @imkiva's review.

We can do follow-up pull requests. If I recall correctly @imkiva is sick yesterday, so if we really want to wait it may take some (more than usual) time

PR #850, opened 2022-12-22
rfc: coercion-agent

SchrodingerZhu:


SchrodingerZhu:
this does not solve the deep coercion yet.

base/src/main/java/org/aya/tyck/tycker/CoercionAgent.java
```diff
@@ -0,0 +1,128 @@
...
+    new Coercer.PathToPi()
+  );
+
+  static public class CoercionFailure extends Throwable {
+    static CoercionFailure INSTANCE = new CoercionFailure();
+
+    @Override
+    public Throwable fillInStackTrace() {
+      return this;
+    }
```
ice1000, line 33: Why?

ice1000 (commented):
Is the throwable object intentionally written like that?

ice1000 (commented):
I also think you don't have to comment out the code. You can just delete it.

SchrodingerZhu:
Close for further development.

Issue #849, opened 2022-12-21
Positivity check is incorrectly implemented


ice1000:
```aya
data Dead : Set
open data Zaoqi (A : Type)
| essentially (A -> Dead)
open data Cat : Type
  | meow : (Zaoqi Cat) → Cat

def catIsDead : Cat → Dead
  | meow (essentially f) => f (meow (essentially f))

def catIsAlive : Cat
  => meow (essentially catIsDead)
  
def catIsBothDeadAndAlive: Dead
  => catIsDead catIsAlive
```

/cc @tsao-chi

ice1000:
https://github.com/JetBrains/Arend/blob/master/base/src/main/java/org/arend/typechecking/covariance/CovarianceChecker.java

HoshinoTented:
😱

tsao-chi:
A more mature implementation would require explicit polarity annotations

References:
https://github.com/agda/agda/issues/2411
https://github.com/agda/agda/pull/6385

PR #851, opened 2022-12-23
WIP: Copy Covariance from Arend(

tsao-chi:
fix https://github.com/aya-prover/aya-dev/issues/849

### DIFF
--- a/base/src/main/java/module-info.java
+++ b/base/src/main/java/module-info.java
@@ -42,4 +42,5 @@ module aya.base {
   exports org.aya.tyck.trace;
   exports org.aya.tyck.unify;
   exports org.aya.tyck;
+    exports org.aya.tyck.covariance;
 }
--- a/base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
+++ b/base/src/main/java/org/aya/concrete/stmt/TeleDecl.java
@@ -3,6 +3,8 @@
 package org.aya.concrete.stmt;
 
 import kala.collection.immutable.ImmutableSeq;
+import kala.collection.mutable.MutableAnySet;
+import kala.collection.mutable.MutableHashSet;
 import kala.collection.mutable.MutableList;
 import kala.control.Either;
 import kala.control.Option;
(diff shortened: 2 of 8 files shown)


ice1000:
I still think the error message is not very helpful. Is there a way to show the exact expression that is problematic?

Ideally, I would appreciate an error message that can help you understand what's wrong in the type checker without debugging the source code of the compiler.

tsao-chi:
> I still think the error message is not very helpful. Is there a way to show the exact expression that is problematic?
> 
> Ideally, I would appreciate an error message that can help you understand what's wrong in the type checker without debugging the source code of the compiler.

I think sourcepos of the ctor definition is enough for user to understand what's wrong in most cases

ice1000:
> > I still think the error message is not very helpful. Is there a way to show the exact expression that is problematic?
> > 
> > Ideally, I would appreciate an error message that can help you understand what's wrong in the type checker without debugging the source code of the compiler.
> 
> I think sourcepos of the ctor definition is enough for user to understand what's wrong in most cases

What if there are 3 mutually recursive data types with each of them 2-10 constructors?

Aya is going to self host and the definition of Term will be inductive-inductively defined together with contexts and type formers. Every introduction and elimination rule is a constructor. This is a real-world example.

tsao-chi:
> What if there are 3 mutually recursive data types with each of them 2-10 constructors?
> 

I think it still should be a very rare case to trigger an error and see a hardly understandable error message.




might implement more detailed error message in future prs

ice1000:
> > What if there are 3 mutually recursive data types with each of them 2-10 constructors?
> > 
> 
> I think it still should be a very rare case to trigger an error and see a hardly understandable error message.
> 
> 
> 
> 
> might implement more detailed error message in future prs

It's exactly when the code gets complicated do people need to look at error messages. So, I would prefer higher quality ones.

ice1000:
I think positivity with the presence of HITs will be more complex. See https://github.com/agda/agda/issues/6336

PR #864, opened 2022-12-29
Simply Refactor PatTycker

HoshinoTented:


base/src/main/java/org/aya/tyck/pat/PatternTycker.java
```diff
@@ -0,0 +1,403 @@
...
+ * A Pattern Tycker for only one use.
+ */
+public final class PatternTycker {
+  public final @NotNull ExprTycker exprTycker;
+  public final @NotNull TypedSubst bodySubst;
+  private final @NotNull TypedSubst sigSubst = new TypedSubst();
+  private @NotNull Def.Signature<?> signature;
+  private @NotNull SeqView<Arg<Pattern>> patterns;
+  private final @NotNull MutableList<Arg<Pat>> wellTyped = MutableList.create();
+  private Term.@UnknownNullability Param currParam = null;
```
HoshinoTented, line 50: I am not sure about this: `private Term.@UnknownNullability Param currParam = null;`

ice1000: What specifically are you not sure about?

ice1000:
Looks good in general

base/src/main/java/org/aya/tyck/pat/PatternTycker.java
HoshinoTented: Maybe we should pass `currParam` by argument? idk!

ice1000: A mutable field is fine. That's the point of OOP! :)

PR #868, opened 2023-01-01
feat: add utils for dynamic congruence class checking

SchrodingerZhu:


SchrodingerZhu:
Why Splay?
-------------------------------------------------------------------

Splay has a good working set property. In type checker, I think it is very likely that we may frequently working on a local set of face lattice, so splay can be a good choice

ice1000 (approved):
💯💯💯💯💯💯

PR #870, opened 2023-01-01
KALA: upgrade to 0.61.0

Glavo:
* Make `kala.base` as multi-release jar;
* For Java 17+, use record to re implement Tuple, allowing use pattern matching for Tuples.


base/src/main/java/org/aya/terck/CallResolver.java
```diff
@@ -5,6 +5,7 @@
 import kala.collection.immutable.ImmutableSeq;
 import kala.collection.mutable.MutableSet;
 import kala.tuple.Tuple;
+import kala.tuple.Tuple2;
```
ice1000, line 8: Is this needed?

ice1000:
bors merge

PR #874, opened 2023-01-06
Reimplement sort system: step 1

ice1000:
Support `piDom` checks.


ice1000:
From 
image
to
image
I think I did my best. Idk why all lines in `Meta.java` are marked as partial hits.

PR #880, opened 2023-01-11
Random refactoring, preparing for fixing `Prop` tycking

ice1000:
These commits have been in my disk for so long so I have made up my mind to create a PR for them


base/src/main/java/org/aya/tyck/StmtTycker.java
```diff
@@ -148,7 +148,7 @@ record Tmp(ImmutableSeq<TeleResult> okTele, Term preresult, Term prebody) {}
       var bodyExpr = fn.body.getLeftValue();
       Term preresult, prebody;
       if (fn.result != null) {
-        preresult = tycker.synthesize(fn.result).wellTyped();
+        preresult = tycker.ty(fn.result);
```
ice1000, line 151: In fact the entire PR is just for this line ;)

PR #881, opened 2023-01-13
Fix sigma projection checking

### LINKED ISSUE #879: Sigma projection check of `Prop` is incorrect
```
> open data Ok Nat : Prop | zero => zo
data Ok (_9 : Nat) : Prop
  | zero ⇒ zo
> zo
zo
> :t zo
Ok zero
> def Sigma => Sig (n : Nat) ** Ok n
def Sigma : Prop ⇒ Σ (n : Nat) ** Ok n
> def test (n : Sigma) => n.2
In file <stdin>:1:24 ->

  1 │   def test (n : Sigma) => n.2
    │                           ╰╯

Error: Unable to project the 2nd element of the expression
         n
       because the type of what you projected on is not a non-Prop Sigma type, but instead:
         Σ (n : Nat) ** Ok n
         (Normalized: Σ (n : Nat) ** Ok n)

> def test (n : Sigma) => n.1
In file <stdin>:1:24 ->

  1 │   def test (n : Sigma) => n.1
    │                           ╰╯

Error: Unable to project the 1st element of the expression
         n
       because the type of what you projected on is not a non-Prop Sigma type, but instead:
         Σ (n : Nat) ** Ok n
         (Normalized: Σ (n : Nat) ** Ok n)
```

ice1000:
Fix #879

See fixed issue for more details

### DIFF
--- a/base/src/main/java/org/aya/tyck/ExprTycker.java
+++ b/base/src/main/java/org/aya/tyck/ExprTycker.java
@@ -147,17 +147,18 @@ public final class ExprTycker extends PropTycker {
         var struct = proj.tup();
         var projectee = instImplicits(synthesize(struct), struct.sourcePos());
         yield proj.ix().fold(ix -> {
-          var pseudoSigma = projectee.type();
+          var pseudoSigma = projectee.type().freezeHoles(state);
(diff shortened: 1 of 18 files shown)


ice1000:
bors r_

PR #884, opened 2023-01-15
Cleanup concrete decls: step 0

ice1000:
This PR:

+ Removes `Decl.Resulted` and `Decl.Telescopic`
+ Make code use `TeleDecl`
+ Introduce `DeclInfo` for good code reuse.


ice1000:
This is good!

image

PR #887, opened 2023-01-18
Refactor the unsafe `ImmutableSeq<String>`

HoshinoTented:


base/src/test/resources/failure/syntax/bad-use-as.aya.txt
```diff
@@ -5,7 +5,21 @@ In file $FILE:6:20 ->
   6 │   open A using (True, False as True)
     │                       ╰───────────╯
 
-Warning: The name `True` introduces ambiguity and can only be accessed through a
-         qualified name
+Warning: The name `True` shadows a previous local definition from outer scope
```
ice1000, line 8: The "previous local definition" does not sound "local" to me 🤔

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -173,7 +173,8 @@ public SeqView<UseHide.Name> useIdsComma(@NotNull GenericNode<?> node) {
       var asId = weakId(useAs.child(WEAK_ID)).data();
       var asAssoc = useAs.peekChild(ASSOC);
       var asBind = useAs.peekChild(BIND_BLOCK);
-      return new UseHide.Name(wholePos, name.data(), asId,
+      var qname = new QualifiedID(wholePos, name.data());    // TODO: Fix this
+      return new UseHide.Name(qname, asId,
```
ice1000, line 177: What's this?

ice1000 (commented):
Can you briefly summerize this PR? I think we should start reading from `ContextUnit`, right? What's the purpose of it?

cli/src/main/java/org/aya/cli/parse/AyaProducer.java
HoshinoTented: 1. The source pos is incorrect
2. I don't update the grammar yet, so `using (Nat::zero as zero)` is still illegal.

base/src/test/resources/failure/syntax/bad-use-as.aya.txt
HoshinoTented: Maybe we need a new `Problem`.

base/src/main/java/org/aya/resolve/context/ModuleSymbol.java
```diff
@@ -0,0 +1,71 @@
...
+
+public interface ModuleSymbol<T extends ContextUnit> {
+
+  /**
+   * Getting the candidates of an unqualified name
+   *
+   * @param unqualifiedName the unqualified name
+   * @return the candidates, probably empty
+   */
+  @NotNull Map<ModulePath, T> getCandidates(@NotNull String unqualifiedName);
```
imkiva, line 23: `resolveUnqualified` looks like a better name: it implies there will be candidates and the input name is unqualified.

base/src/main/java/org/aya/resolve/context/ModuleSymbol.java
```diff
@@ -0,0 +1,71 @@
...
+   * Trying to get a symbol of an optional component and an unqualified name.
+   *
+   * @param component       an optional component, none if `This`
+   * @param unqualifiedName the unqualified name
+   */
+  @NotNull Result<T, Error> getDefinitely(@NotNull ModulePath component, @NotNull String unqualifiedName);
+
+  boolean contains(@NotNull String unqualified);
+
+  boolean containsDefinitely(@NotNull String unqualified);
```
imkiva, line 51: What are the differences of these methods?

Possible to implement with `getCandidates().isNotEmpty()`?

base/src/main/java/org/aya/resolve/context/ModuleSymbol.java
```diff
@@ -0,0 +1,71 @@
...
+  @NotNull Map<ModulePath, T> getCandidates(@NotNull String unqualifiedName);
+
+  /**
+   * Getting a symbol of an unqualified name {@param unqualifiedName} in component {@param component}
+   *
+   * @param component       the component
+   * @param unqualifiedName the unqualified name
+   * @return none if not found
+   */
+  @NotNull Option<T> getQualifiedMaybe(@NotNull ModulePath component, @NotNull String unqualifiedName);
```
imkiva, line 32: Possible to implement with `getCandidates().getOption()`?

imkiva (commented):
1. What does "Definitely" mean in those methods whose names end with that?

2. Is it possible to merge `ContextUnit` and `GlobalSymbol`?

3. What is the purpose of `MockModuleContext`? If I read the code correctly, this context is for data/struct definitions, and they cannot import other modules, right? 
I don't think it is necessary to tell modules from definitions because we may want to import/export something in the impl-block.

4. Is it possible to implement "getXXXMaybe"/"getXXXFefinitely" with "getCandidates" in `ModuleSymbol`?

imkiva:
For 1, @ice1000 told me that "definitely" means a success resolve without confusion. In this case, remove all suffixes. 

Rationale: In `ModuleContext`, we add "maybe" suffix for a fail-able resolve. No suffix always means a success and unambiguous resolve. We should keep method names consistent.

base/src/main/java/org/aya/resolve/context/ModuleContext.java
```diff
@@ -1,155 +1,226 @@
...
   }
 
+  /**
+   * Importing one module export.
+   *
+   * @param accessibility of importing, re-export if public
+   * @param componentName the name of the module
+   * @param moduleExport  the module
+   */
   default void importModule(
```
imkiva, line 71: `importModuleExports` imports everything from a module according to your comment.

`importModule` imports only exported things from a module.

I am confused by their names.

base/src/main/java/org/aya/resolve/context/ContextUnit.java
```diff
@@ -0,0 +1,58 @@
...
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.resolve.context;
+
+import org.aya.concrete.stmt.Stmt;
+import org.aya.ref.AnyVar;
+import org.aya.ref.DefVar;
+import org.aya.ref.LocalVar;
+import org.jetbrains.annotations.NotNull;
+
+public sealed interface ContextUnit {
```
imkiva, line 11: In this class we track where a symbol comes from. 
But it seems we don't check if a symbol is usable by judging its home?

imkiva: Okay, I see two instanceof checks of `ContextUnit.Exportable`, in `GlobalSymbol` and `ModuleLikeContext`.

Does this mean we should only track the accessibility?

HoshinoTented: They track the accessibility and limit the type of var:
* We can only export/import `DefVar`s
* Only `DefVar`s have accessibility (I won't to see a `public GeneralizedVar`)

HoshinoTented:
For 2:
Another purpose of `MockModuleContext` is to permit some private symbols to be used in the signature. However, we can add a new field in `PhysicalModuleContext`

base/src/main/java/org/aya/resolve/context/ContextUnit.java
imkiva: Do you think we need variants other than `Exportable` and `NotExportable`? From the current code, there is only one non-trivial usage of `ContextUnit` and it is just instanceof check of `Exportable`. I guess we only need these two: one for all DefVars with accessibility, and the other for the rest cases.

Correct me if there are other design considerations.

HoshinoTented: Then a `LocalVar` **is able to** be added to ModuleSymbol which is weird.
These variants prevent the programmer write stupid code.

base/src/main/java/org/aya/resolve/context/PhysicalModuleContext.java
```diff
@@ -43,6 +39,17 @@ public PhysicalModuleContext(@NotNull Context parent, @NotNull ImmutableSeq<Stri
...
 
+  @Override
+  public @Nullable ContextUnit getUnqualifiedLocalMaybe(
+    @NotNull String name,
+    Stmt.@Nullable Accessibility accessibility,
+    @NotNull SourcePos sourcePos
+  ) {
+    var permitted = permits.getOption(name);
+    if (permitted.isDefined()) return permitted.get();
+    return ModuleContext.super.getUnqualifiedLocalMaybe(name, accessibility, sourcePos);
```
imkiva, line 50: Why we need this `permits`?

imkiva: I guess the `permits` is for the code below?
```
open data N | zero | suc N
open data N | zero | suc N
```

base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
```diff
@@ -84,14 +87,21 @@ public void enterBody() {
     return new ExprResolver(ctx, RESTRICTIVE, allowedGeneralizes, reference, MutableStack.create(), this::addReference);
   }
 
+  private @Nullable Stmt.Accessibility accessibility() {
+    return switch (where.peek()) {
+      case Where.Body body -> null;
+      case Where.Head head -> head.accessibility();
+    };
+  }
+
```
imkiva, line 96: Why need this? The `accessibility` is the accessibility of the currently being-resolved definition. It has no effect on `checkAccessibility`.

image

base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
```diff
@@ -144,13 +154,13 @@ public void enterBody() {
         },
         right -> right.descent(this)
       ));
-      case Expr.Unresolved(var pos, var name) -> switch (ctx.get(name)) {
+      case Expr.Unresolved(var pos, var name) -> switch (ctx.get(name, accessibility()).data()) {
```
imkiva, line 157: ditto

base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
imkiva: I even believe the check of accessibility should be done in `ExprResolver` rather than here.

base/src/main/java/org/aya/resolve/context/PhysicalModuleContext.java
HoshinoTented: It is for the code:
```
private data N | zero | suc N
```

base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
HoshinoTented: > Why need this? The `accessibility` is the accessibility of the currently being-resolved definition. It has no effect on `checkAccessibility`.

`Context#get` -> `Context#getUnqualified` -> `Context#getUnqualifiedLocal` -> `Context#getUnqualifiedLocalMaybe` -> `Context#checkAccessibility`

imkiva: This looks scary and it needs some hacking in PhysicalModuleContext.java (the `permits`). What will go wrong if we put the accessibility check just in `ExprResolver` after `ctx.get(name).data()`

base/src/main/java/org/aya/resolve/context/PhysicalModuleContext.java
imkiva: emmm

imkiva: That's because the context is checking the accessibility without knowing the being-resolved definition. Right?

base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
imkiva: The impl of `checkAccessibility` always gets called after every unqualified resolve, and it says "public definitions cannot refer to private ones". There are problems:
- What should I do when I just want to hide the detail to other modules?
- Not all places need to check accessibility, for example, as you said: `private data N | zero | suc N`

base/src/main/java/org/aya/resolve/context/PhysicalModuleContext.java
HoshinoTented: Because `data N` is private. However, I removed the accessibility checking.

ice1000: I still think checking acc is useful. If only there's a simple way to achieve that....

HoshinoTented: The way I impl it is too dirty, I have no better idea.

ice1000: Okay. We can save it for later

PR #888, opened 2023-01-21
Gradle 8.0-rc-2

ice1000:
This is a slightly aggressive update, but I think 8.0 has done some good upgrades that are good to have.


ice1000:
888!

PR #891, opened 2023-01-28
Improve scripts

ice1000:
This PR resolves a warning in the nightly build workflow.

bors r+


ice1000:
This does not actually resolve it as there are more usages of the same config... My bad on merging it too quickly.

PR #894, opened 2023-01-31
Refactor qid refactor round two

ice1000:
Further refactor #887 after #893 

/cc @imkiva @HoshinoTented

### DIFF
--- a/base/src/main/java/org/aya/concrete/stmt/QualifiedID.java
+++ b/base/src/main/java/org/aya/concrete/stmt/QualifiedID.java
@@ -38,10 +38,6 @@ public record QualifiedID(
     return join(ids());
   }
 
-  public @NotNull ModulePath.Qualified asModulePath() {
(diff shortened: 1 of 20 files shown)


base/src/main/java/org/aya/concrete/stmt/QualifiedID.java
```diff
@@ -38,10 +38,6 @@ public boolean isUnqualified() {
     return join(ids());
   }
 
-  public @NotNull ModulePath.Qualified asModulePath() {
```
HoshinoTented, line 41: I think I need this method.

ice1000: In general we should eliminate it. Where do you need it?

HoshinoTented: Forget what I said 🥲

PR #898, opened 2023-02-09
Implement #896

### LINKED ISSUE #896: Support projection out of conCalls
Basically this: image

The feature is analogous to projecting from a tuple. It's good because:

1. Otherwise you'll need dumb pattern matching to extract the fields
2. Usually you can just write out the field directly because _you do have that piece of information_ but why not
3. This syntax is not ambiguous to anything present or potentially useful in the future
4. It will not check for Prop-ness because, again, _you do have that piece of information_.

ice1000:
Fix #896


ice1000:
This is just a simpler idea I came up with long time ago. I used this feature as an excuse to cherry-pick something from the ongoing class overhaul branch

PR #899, opened 2023-02-09
Use snapshot jacoco

ice1000:
We need to switch to the release version immediately once 0.8.9 is released and delete the extra maven repository. I conjecture that this can improve the test coverage drastically.


ice1000:
....

ice1000:
Let's wait until 0.8.9 release then.

ice1000:
It increased by a lot locally though. I guess codecov has its own algorithm for calculation

PR #902, opened 2023-02-11
Improve LaTeX literate backend

imkiva:
/cc @ice1000

### DIFF
--- a/cli-impl/src/main/java/org/aya/cli/literate/FaithfulPrettier.java
+++ b/cli-impl/src/main/java/org/aya/cli/literate/FaithfulPrettier.java
@@ -1,4 +1,4 @@
-// Copyright (c) 2020-2022 Tesla (Yinsen) Zhang.
+// Copyright (c) 2020-2023 Tesla (Yinsen) Zhang.
 // Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
 package org.aya.cli.literate;
 
@@ -9,8 +9,8 @@ import kala.collection.mutable.MutableList;
 import kala.text.StringSlice;
 import kala.tuple.Tuple;
 import kala.tuple.Tuple4;
-import org.aya.prettier.BasePrettier;
 import org.aya.generic.AyaDocile;
+import org.aya.prettier.BasePrettier;
 import org.aya.pretty.doc.Doc;
 import org.aya.util.error.SourcePos;
 import org.aya.util.prettier.PrettierOptions;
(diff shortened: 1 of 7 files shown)


cli-impl/src/main/java/org/aya/cli/literate/FaithfulPrettier.java
```diff
@@ -59,19 +49,25 @@ private static void checkHighlights(@NotNull SeqLike<HighlightInfo> highlights)
...
+      var knifeCut = twoKnifeThreeParts(raw, base, current.sourcePos());
+      // move forward
+      raw = knifeCut.remaining;
+      base = knifeCut.base;
+
+      // If there's orphan text before the highlighted cut, add it to result as plain text.
+      if (!knifeCut.before.isEmpty()) {
+        // TODO: handle whitespaces in the lexer, and use a new highlight type for them.
+        //  this workaround solution does not work for whitespace in LaTeX.
+        var lines = knifeCut.before.toString().split("\n", -1);
```
imkiva, line 63: ij-parsing-core seems not providing detailed whitespace kind, e.g. tabs, spaces, line breaks, etc.

ice1000: Whitespaces can be ~ in LaTeX, and we are not using tabs for now

ice1000: I think it's related to how our lexer is written. We are probably not distinguishing among various whitespace tokens

ice1000 (approved):
I'm convinced

ice1000:
bors merge

pretty/src/main/java/org/aya/pretty/backend/latex/DocTeXPrinter.java
```diff
@@ -40,13 +41,31 @@ public class DocTeXPrinter extends StringPrinter<DocTeXPrinter.Config> {
...
+    id("(|"), id("|)"),
+    Tuple.of("\u2987", "(|"),
+    Tuple.of("\u2988", "|)"),
     id("[|"), id("|]"),
+    Tuple.of("\u27E6", "[|"),
+    Tuple.of("\u27E7", "|]"),
+    Tuple.of("{|", "\\{|"),
+    Tuple.of("|}", "|\\}"),
+    Tuple.of("\u2983", "\\{|"),
+    Tuple.of("\u2984", "|\\}"),
```
ice1000, line 66: I think we should use dedicated LaTeX symbols. For instance I guess we have `\lBrace` `\rBrace` something like dat

Issue #907, opened 2023-02-13
Predicativity of extension types (obviously) for terck


ice1000:
Background:

Agda's termination checker assumes predicativity, so it allows, e.g. f (cons g) = f (g blah) and this is considered okay. This has a consequence: Prop has to be predicative, and I think it sucks.
So, we accidentally (and intentionally, after we've realized this) didn't implement this.

Observation:

The recursion principle of set truncation is attached (instead of code block, because with inlay hints it helps understanding the types of the variables). The recursive call rec f x (p k) actually requires some sense of predicativity (but limited to paths) because one needs to consider p k smaller than trunc a b p q i j, where k is a local variable. This implies that we need some sort of predicaitivity assumptions in the termination checker whatsoever.

![image](https://user-images.githubusercontent.com/16398479/218397403-c64b2e96-6966-4b54-9831-a3f592eaf344.png)

Agda has this by default so they were probably not bogged down by this, but if anyone wants to do cubical with pattern matching on HITs, this is a problem to be taken care of.

ice1000:
I have posted this to AK Jr.'s discord: https://discord.com/channels/767397347218423858/767397347218423861/1074568251029327913

ice1000:
Update: this should actually work for pattern matching of all types in `Type` or `Set`. It should be forbidden for types in `Prop`.

ice1000:
Further predicativity assumption is left as future work, fixed by #908

PR #908, opened 2023-02-13
Predicativity assumption for terck

ice1000:
This partially addresses #907. There are two issues left:

+ Unfold some simple decls before tercking. This cannot be done because `TermConsumer` does not allow blocking a sub-tree traversal. /cc @wsx-ucb 
+ Predicativity assumption for `Type`


ice1000:
bors merge

Issue #719, opened 2022-11-23
`let open` for local openings?


imkiva:
Extend the `let` expression with local opening, like

```aya
let open Nat in
suc zero
```

ice1000:
Fixed by #909

PR #909, opened 2023-02-13
Let Open Expr

HoshinoTented:
Solve #719


base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
```diff
@@ -195,6 +198,18 @@ public void enterBody() {
           newBody
         );
       }
+      case Expr.LetOpen(var pos, var component, var useHide, var body) -> {
+        var innerCtx = new NoExportContext(ctx);
+        // open module
+        innerCtx.openModule(
```
HoshinoTented, line 204: Maybe a `openModule(Command.Open)` ?

PR #915, opened 2023-02-14
Rework `Term` visitor to include patterns.

wsx-ucb:
Yak shaving.


base/src/main/java/org/aya/core/def/CtorDef.java
```diff
@@ -45,6 +46,16 @@ public CtorDef(
     return ref;
   }
 
+  public @NotNull CtorDef update(@NotNull ImmutableSeq<Arg<Pat>> pats, @NotNull ImmutableSeq<Term.Param> selfTele, @NotNull Term result, @NotNull Partial.Split<Term> clauses) {
+    return pats.sameElements(this.pats, true) && selfTele.sameElements(this.selfTele, true) && result == this.result && clauses == this.clauses
+      ? this : new CtorDef(dataRef, ref, pats, ownerTele, selfTele, clauses, result, coerce);
```
ice1000, line 51: You should never `new CtorDef`! In the constructor of this class, we mutate `ref.core` so the old one will never be referred to

ice1000 (changes requested):

base/src/main/java/org/aya/core/def/CtorDef.java
wsx-ucb: Will this cause any problem?
I think the updated `CtorDef` is still well-formed, right?
(In the sense that for `updated = old.update(...)`, `updated.ref.core == updated`.)

wsx-ucb: Oh I see it now, the reference of all other `DefVar` will now point to the updated definition, which is undesirable.
The same applies to `FnDef` etc, right?

wsx-ucb: Updated the way we visit `GenericDef` to only allow `Consumer`, and no transformations.

ice1000 (approved):

ice1000:
bors merge

PR #916, opened 2023-02-14
Unfold before termination check!

wsx-ucb:
Really not sure what is the best solution here.
The current implementation is very hacky...
Based on #915.

### DIFF
--- a/base/src/main/java/org/aya/terck/CallResolver.java
+++ b/base/src/main/java/org/aya/terck/CallResolver.java
@@ -153,7 +153,7 @@ public record CallResolver(
 
   @Override public @NotNull Term pre(@NotNull Term term) {
     // TODO: Rework error reporting to include the original call
-    term = new Expander.ConservativeWHNFer(new TyckState(factory), ImmutableSet.of(caller.ref)).apply(term);
+    term = new Expander.ConservativeWHNFer(new TyckState(factory), ImmutableSet.from(targets.map(Def::ref))).apply(term);
     if (term instanceof Callable call) resolveCall(call);
     return DefVisitor.super.pre(term);
   }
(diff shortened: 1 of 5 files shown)


ice1000 (commented):
Could you please also update the stdlib with `pmap`?

wsx-ucb:
Updated the `pmap` usage.
I also attempted to switch to using `pmapd` in `Data/Vec.aya`, but it seems to lead to an internal exception that is unrelated to `terck`.
I marked this later usage as a todo.

base/src/main/java/org/aya/terck/CallResolver.java
```diff
@@ -151,7 +153,7 @@ private Relation compareConArgs(@NotNull ImmutableSeq<Arg<Term>> conArgs, @NotNu
 
   @Override public @NotNull Term pre(@NotNull Term term) {
     // TODO: Rework error reporting to include the original call
-    // term = whnf(term);
+    term = new Expander.ConservativeWHNFer(new TyckState(factory), ImmutableSet.of(caller.ref)).apply(term);
```
ice1000, line 156: Should it be `targets` instead of `caller.ref`?

wsx-ucb: Oh good point, in principle we should avoid unfolding any recursive function in the current SCC, and `target` contains all of them, right?

ice1000: I think so. Cc @imkiva

ice1000:
Could you please also replace set truncation's eliminator with pmap?

ice1000 (approved):
bors d=wsx-ucb

imkiva (approved):
bors d=wsx-ucb

ice1000:
bors merge

PR #925, opened 2023-02-18
Anonymous examples

### LINKED ISSUE #897: Anonymous examples
We should allow `example def [: xxx] => xxx`

HoshinoTented:
Resolves #897

### DIFF
--- a/base/src/main/java/org/aya/resolve/visitor/StmtShallowResolver.java
+++ b/base/src/main/java/org/aya/resolve/visitor/StmtShallowResolver.java
@@ -173,7 +173,12 @@ public record StmtShallowResolver(@NotNull ModuleLoader loader, @NotNull Resolve
     };
     decl.setCtx(ctx);
     decl.ref().module = ctx.moduleName();
-    ctx.defineSymbol(decl.ref(), decl.accessibility(), decl.sourcePos());
+
+    // TODO: this is too specific!! is it good?
+    if (!(decl instanceof TeleDecl.FnDecl fnDecl) || (!fnDecl.isAnonymous)) {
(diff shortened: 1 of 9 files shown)


base/src/main/java/org/aya/resolve/visitor/StmtShallowResolver.java
```diff
@@ -173,7 +173,12 @@ private void resolveOpInfo(@NotNull Decl decl, @NotNull Context context) {
     };
     decl.setCtx(ctx);
     decl.ref().module = ctx.moduleName();
-    ctx.defineSymbol(decl.ref(), decl.accessibility(), decl.sourcePos());
+
+    // TODO: this is too specific!! is it good?
```
ice1000, line 177: I think it's fine

cli-impl/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -250,58 +250,74 @@ public SeqView<UseHide.Name> useIdsComma(@NotNull GenericNode<?> node) {
...
+          self.error(node.childrenView().first(), "Expect a name");
+          return null;
+        } else {
+          // do generate
+          return Constants.randomName(info.sourcePos());
+        }
+      } else {
+        return name;
+      }
+    }
```
imkiva, line 271:
```java
if (name != null) return name;
if (require) return self.error(node.childrenView().first(), "Expect a name");
return Constants.randomName(info.sourcePos());
```
and
```java
private <T> T error(@NotNull GenericNode<?> node, @NotNull String message) {
    reporter.report(new ParseError(sourcePosOf(node), message));
    return null;
}
```

imkiva:
bors merge

PR #929, opened 2023-02-21
Fix some potential bugs

HoshinoTented:
- The keys of `ResolveInfo#imports` and `ResolveInfo#reExports` are now `ModulePath.Qualified` as they should.
- Supports module renaming in `CompiledAya` to fix a TODO.

### DIFF
--- a/base/src/main/java/org/aya/core/serde/CompiledAya.java
+++ b/base/src/main/java/org/aya/core/serde/CompiledAya.java
@@ -51,7 +51,7 @@ public record CompiledAya(
   /**
    * @param rename not empty
    */
-  record SerImport(@NotNull ImmutableSeq<String> moduleName, @NotNull ImmutableSeq<String> rename,
+  record SerImport(@NotNull ModulePath.Qualified moduleName, @NotNull ImmutableSeq<String> rename,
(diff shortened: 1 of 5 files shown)


base/src/main/java/org/aya/core/serde/CompiledAya.java
```diff
@@ -51,7 +51,7 @@ public record CompiledAya(
   /**
    * @param rename not empty
    */
-  record SerImport(@NotNull ImmutableSeq<String> moduleName, @NotNull ImmutableSeq<String> rename,
+  record SerImport(@NotNull ModulePath.Qualified moduleName, @NotNull ImmutableSeq<String> rename,
```
HoshinoTented, line 54: `ModulePath.Qualified` represents a **name** in a context instead of an absolute path. In fact, I think it should be renamed `ModuleName`, and `ModulePath` for absolute paths.

ice1000: Oh okay.

ice1000: Can you do this?

PR #429, opened 2022-06-30
Struct WIP

mio-19:
resolves https://github.com/aya-prover/aya-dev/issues/386

PR #419, opened 2022-05-22
Struct WIP

tsao-chi:
resolves https://github.com/aya-prover/aya-dev/issues/386

PR #931, opened 2023-02-22
Class system overhaul

### LINKED ISSUE #386: Overhaul struct

ice1000:
Fix #386, close #429 and close #419


ice1000:
Let's close this for now, and I'll continue to work on this soon.

bors r+

PR #932, opened 2023-02-22
Literate: KaTeX, ServerSideRendering, and options to inline highlight styles

imkiva:
After merging this PR, please add `--pretty-ssr` to [aya-prover-docs](https://github.com/aya-prover/aya-prover-docs/blob/711029dffc6291c51f85756cbc2f7ca3f2fe5516/scripts/build-aya.js#L32).


ice1000 (approved):
bors r+

PR #933, opened 2023-02-23
Refactorings, fix tyck ordering in patterns, improve latex backend, add things to library

### LINKED ISSUE #930: We should use `enum X { Obj; }` for singletons!
This sounds like the best way of using singletons in Java

ice1000:
Fix #930


imkiva (approved):
bors merge

PR #935, opened 2023-02-26
ModuleName <> ModulePath

HoshinoTented:
* ModuleName: a local name used in a local scope, such as a module context.
* ModulePath: an absolute path to a module.

### DIFF
--- a/base/src/main/java/org/aya/core/term/MetaTerm.java
+++ b/base/src/main/java/org/aya/core/term/MetaTerm.java
@@ -19,6 +19,10 @@ public record MetaTerm(
   @NotNull ImmutableSeq<@NotNull Arg<@NotNull Term>> contextArgs,
   @NotNull ImmutableSeq<@NotNull Arg<@NotNull Term>> args
 ) implements Callable {
+  public MetaTerm {
+
+  }
+
   public @NotNull MetaTerm update(@NotNull ImmutableSeq<Arg<Term>> contextArgs, @NotNull ImmutableSeq<Arg<Term>> args) {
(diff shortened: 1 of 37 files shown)


base/src/main/java/org/aya/core/term/MetaTerm.java
```diff
@@ -19,6 +19,10 @@ public record MetaTerm(
   @NotNull ImmutableSeq<@NotNull Arg<@NotNull Term>> contextArgs,
   @NotNull ImmutableSeq<@NotNull Arg<@NotNull Term>> args
 ) implements Callable {
+  public MetaTerm {
+
+  }
+
```
ice1000, line 25: What are these

base/src/main/java/org/aya/resolve/context/ModuleName.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override public @NotNull Qualified resolve(@NotNull String name) {
+      return new Qualified(ids.appended(name));
+    }
+
+    @Override public @NotNull Qualified concat(@NotNull ModuleName path) {
+      return new Qualified(ids.concat(path.ids()));
+    }
+
+    @Override public @NotNull String toString() {
+      return ids.joinToString(Constants.SCOPE_SEPARATOR);
```
ice1000, line 64: QualifiedID.join

ice1000 (approved):
With minor nitpicking

base/src/main/java/org/aya/core/term/MetaTerm.java
HoshinoTented: 😱 IDK!!

PR #943, opened 2023-03-04
More Stdlib + Misc

HoshinoTented:
* `Reflect` (Coq style)
* `=? : Nat -> Nat -> Bool` and its properties
* more lemmas for `Bool`
* an error for `overlap def foo => bar`

The last change looks lonely if I create a new PR for it, so I put them together.

### DIFF
--- a/base/src/test/resources/success/common/src/Arith/Nat/Core.aya
+++ b/base/src/test/resources/success/common/src/Arith/Nat/Core.aya
@@ -1,6 +1,9 @@
 open import Paths
 open import Algebra::Formula
 
+import Arith::Bool::Core
+open   Arith::Bool::Core
+
 open data Nat
 | zero
 | suc Nat
@@ -60,3 +63,10 @@ private overlap def *-assoc-lemma : Associative (*)
 | 0, _, _ => idp
 | suc m, _, _ => <=3=> (+-*-distl _) (pmap (+ _) *-assoc) +-comm
 inline def *-assoc {m n k : Nat} => *-assoc-lemma m n k
+
+overlap def infix =? (m n : Nat) : Bool
+| 0    , 0     => true
+| suc _, 0     => false
+| 0    , suc _ => false
+| suc m, suc n => m =? n
+looser + *
--- a/base/src/test/resources/success/common/src/Arith/Nat/Core.aya
+++ b/base/src/test/resources/success/common/src/Arith/Nat/Core.aya
@@ -1,8 +1,7 @@
 open import Paths
 open import Algebra::Formula
 
-import Arith::Bool::Core
-open   Arith::Bool::Core
+open import Arith::Bool::Core
 
 open data Nat
 | zero
--- a/base/src/test/resources/success/common/src/Arith/Nat/Core.aya
+++ b/base/src/test/resources/success/common/src/Arith/Nat/Core.aya
@@ -1,8 +1,6 @@
 open import Paths
 open import Algebra::Formula
 
-open import Arith::Bool::Core
-
 open data Nat
 | zero
 | suc Nat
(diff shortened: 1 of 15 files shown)


ice1000:
It's better to define =? as decEq which returns decidable. In fact I think it is in the library

base/src/test/resources/success/common/src/Arith/Nat/Properties.aya
```diff
@@ -10,3 +16,42 @@ def z≠s {a : Nat} (p : 0 = suc a) : False => (\i => diag (p i)).coe a
...
+
+def =?-= {m n : Nat} (p : (m =? n) = true) : m = n
+| {0}, {0}, _ => idp
+| {0}, {suc n'}, p => elim (t≠f (sym p))
+| {suc m'}, {0}, p => elim (t≠f (sym p))
+| {suc m'}, {suc n'}, p =>
+  // p : (suc m' =? suc n') = true
+  //   : (m' =? n') = true            simpl.
+  // Goal : suc m' = suc n'
+  pmap suc (=?-= {m'} {n'} p)
```
ice1000, line 40: Is it possible to do it w/o these implicits?

base/src/test/resources/success/common/src/Arith/Nat/Core.aya
```diff
@@ -60,3 +63,10 @@ private overlap def *-assoc-lemma : Associative (*)
 | 0, _, _ => idp
 | suc m, _, _ => <=3=> (+-*-distl _) (pmap (+ _) *-assoc) +-comm
 inline def *-assoc {m n k : Nat} => *-assoc-lemma m n k
+
+overlap def infix =? (m n : Nat) : Bool
```
ice1000, line 67: This definition is not overlapping. I'll try to write a warning that detects such non-overlapping `overlap` definitions.

HoshinoTented:
> It's better to define =? as decEq which returns decidable. In fact I think it is in the library

What is the math symbol of `decEq`? I am not friendly with the Agda-Style library.

base/src/test/resources/success/common/src/Arith/Nat/Properties.aya
HoshinoTented: 🤔what is `w/o`?

ice1000:
> > It's better to define =? as decEq which returns decidable. In fact I think it is in the library
> 
> What is the math symbol for `decEq`? I am not friendly with the Agda-Style library.

You don't have to always use unicode, we can use English names

base/src/test/resources/success/common/src/Arith/Nat/Equality.aya
```diff
@@ -0,0 +1,20 @@
...
+open import Arith::Nat::Properties
+open import Logic::Dec
+open import Logic::False
+
+def ≐-suc {m n : Nat} (p : Dec (m = n)) : Dec (suc m = suc n)
+| yes p => yes (pmap suc p)
+| no np => no (\ p =>
+  // p  : suc m = suc n
+  // np : m = n -> False
+  // Goal : False
```
ice1000, line 12: Are these information available in vscode?

base/src/test/resources/success/common/src/Arith/Nat/Properties.aya
ice1000: It's 'without'

base/src/test/resources/success/common/src/Arith/Nat/Equality.aya
HoshinoTented: No, I list them manually for me.

ice1000: I think they are

ice1000: Use a hole, `{??}` as an expression

base/src/test/resources/success/common/src/Arith/Nat/Properties.aya
HoshinoTented: These comments can be deleted. But it helps someone who reads this code.

ice1000:
Good
bors r+

PR #947, opened 2023-03-30
Support LaTeX math formula in literate mode

imkiva:
Example:
```md
Inline math: $\mathrm{type}$.
Math block:

$$
\LaTeX
$$
```

This works with markdown processors that support KaTeX (like Vuepress/Vitepress)

### DIFF
--- a/base/src/main/java/module-info.java
+++ b/base/src/main/java/module-info.java
@@ -13,6 +13,8 @@ module aya.base {
   exports org.aya.concrete.desugar;
   exports org.aya.concrete.error;
   exports org.aya.concrete.remark;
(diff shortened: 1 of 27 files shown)


ice1000 (commented):
Why is math block always `align*`?

imkiva:
> Why is math block always `align*`?

I don't know what to use to be honest. Maybe just the math environment?

ice1000:
> > Why is math block always `align*`?
> 
> I don't know what to use to be honest. Maybe just the `math` environment?

`\[` and `\]` sounds good.

imkiva:
bors merge

PR #948, opened 2023-03-31
Show compiler diagnostics in literate output

imkiva:
This is the best feature I've ever implemented! The JavaScript code used to show the popup is copied from [1lab](https://github.com/plt-amy/1lab/blob/5e5a22abce8a5cfb62b5f815e1231c1e34bb0a12/support/web/js/highlight-hover.ts#L22) (thanks!!!!).

image


ice1000:
In the docs, we should make the hover window less dark (or maybe make it white-ish).

ice1000:
merge?

imkiva:
> merge?

wait a sec, lots of tweaks are coming ~

cli-impl/src/main/java/org/aya/cli/literate/FaithfulPrettier.java
```diff
@@ -27,6 +29,30 @@ private static void checkHighlights(@NotNull ImmutableSeq<HighlightInfo> highlig
     });
   }
 
+  private static @NotNull ImmutableSeq<HighlightInfo> merge(
```
HoshinoTented, line 32: Maybe move to `SyntaxHighlight`? It manipulates `HighlightInfo`, but `FaithfulPrettier` doesn't.

imkiva: See the updated faithful prettier, it now manages all highlights and problems of a single file, rather than the corresponding code part.

ice1000:
Merge?

ice1000:
JavaScript master....

imkiva:
bors merge

PR #955, opened 2023-04-06
Resolver should not add empty modules when no write access is needed

imkiva:
this is the root cause of #954


ice1000:
Does this take over #954?

HoshinoTented (dismissed):
Amazing

HoshinoTented:
> Does this take over #954?

We still need #954. Whatever the candidates, the rhs (bors) always overwrites the corresponding records in lhs (this).

imkiva:
> Does this take over #954?

No, we need both

imkiva:
bors d=HoshinoTented

PR #957, opened 2023-04-11
Upgrade upstream

ice1000:
This PR replaces `MarkerNode::text` with a dynamically computed string slice.


cli-impl/src/main/java/org/aya/cli/parse/AyaProducer.java
```diff
@@ -76,6 +76,7 @@
  * @see AyaPsiElementTypes
  */
 public record AyaProducer(
+  @NotNull String code,
```
imkiva, line 79: This seems unused

ice1000: Qs, my bad, feel free to delete

PR #959, opened 2023-04-15
Literate library

imkiva:
# Summary

This PR adds support for pretty printing a library in our literate way. To customize the generation, new fields were added to `aya.json`:

```json
"literate": {
  "linkPrefix": "/aya-prover/base/src/test/resources/success/build/pretty/",
  "pretty": {
    "prettierOptions": {
      "map": {
        "InlineMetas": true,
        "ShowImplicitArgs": false,
        "ShowImplicitPats": true,
        "ShowLambdaTypes": false
      }
    },
    "renderOptions": {
      "colorScheme": "Emacs",
      "styleFamily": "Default"
    }
  }
},
```

- `linkPrefix`: the prefix to the link of "go to definition". This is useful when the generated page is not in the root folder of the HTTP server. The default value is "/" if unspecified.
- `pretty`: same as the `prettierOptions` and `renderOptions` in `~/.aya/repl_config.json`. The default value comes from `repl_config.json` if unspecified.

# How to build a literate library?

Take the library `base/src/test/resources/success/` for example:
```shell
aya --make base/src/test/resources/success --pretty-stage literate --pretty-format html
```

The above command outputs HTML files which are always located in `<library-root>/build/pretty`. Much like how we compile a single literate file.
(Remember the `"linkPrefix": "/aya-prover/base/src/test/resources/success/build/pretty/"` above?)

Note that the command line options `--prettyDir` and `--output` has no effect on libraries.

## For developers
[LibraryTest.testLiterate](https://github.com/aya-prover/aya-dev/blob/6ce530cddf2256901b02edfc5fb460975d393fc9/base/src/test/java/org/aya/test/LibraryTest.java#L46)

# TODO

- [ ] Collect compiler diagnostics for libraries, see #948 
- [ ] Build highlights for `import`, `open`

### DIFF
--- a/base/src/main/java/org/aya/concrete/visitor/StmtFolder.java
+++ b/base/src/main/java/org/aya/concrete/visitor/StmtFolder.java
@@ -49,7 +49,7 @@ public interface StmtFolder<R> extends Function<Stmt, R>, ExprFolder<R> {
     return switch (stmt) {
       case Generalize g -> g.variables.foldLeft(acc, (a, v) -> foldVarDecl(a, v, v.sourcePos, noType()));
       case Command.Module m -> foldModuleDecl(acc, m.sourcePos(), new ModuleName.Qualified(m.name()));
-      case Command.Import i -> acc; // we are no need to fold `i.path()`
+      case Command.Import i -> foldModuleRef(acc, i.sourcePos(), i.path().asName());
       case Command.Open o when o.fromSugar() -> acc;  // handled in `case Decl` or `case Command.Import`
       case Command.Open o -> {
         acc = foldModuleRef(acc, o.sourcePos(), o.path());
--- a/base/src/main/java/org/aya/concrete/Expr.java
+++ b/base/src/main/java/org/aya/concrete/Expr.java
@@ -652,6 +652,19 @@ public sealed interface Expr extends AyaDocile, SourceNode, Restr.TermLike<Expr>
       }
     }
 
+    public record ListCompNames(
+      @NotNull Expr monadBind,
+      @NotNull Expr functorPure
+    ) {
+      public ListCompNames fmap(@NotNull Function<Expr, Expr> f) {
+        return new ListCompNames(f.apply(monadBind), f.apply(functorPure));
+      }
+
+      public boolean identical(@NotNull ListCompNames names) {
+        return monadBind == names.monadBind && functorPure == names.functorPure;
+      }
+    }
+
     /**
      * <h1>Array Comp(?)</h1>
      * <p>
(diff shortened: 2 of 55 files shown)


ice1000:
What is the semantics of `fileMod`?

imkiva:
> What is the semantics of `fileMod`?

The module name up to file. 
```aya
// A/B.aya
data X | x
```

The constructor `x` lives in
- module `A::B::X`
- file module `A::B`

base/src/main/java/org/aya/concrete/visitor/StmtFolder.java
```diff
@@ -49,7 +49,7 @@
     return switch (stmt) {
       case Generalize g -> g.variables.foldLeft(acc, (a, v) -> foldVarDecl(a, v, v.sourcePos, noType()));
       case Command.Module m -> foldModuleDecl(acc, m.sourcePos(), new ModuleName.Qualified(m.name()));
-      case Command.Import i -> acc; // we are no need to fold `i.path()`
+      case Command.Import i -> foldModuleRef(acc, i.sourcePos(), i.path().asName());
```
HoshinoTented, line 52: Do we need to `fold` the `ModuleName`s which `import` introduces? For example, `import Data::Nat as Nat` introduces a ModuleName `Nat`.

imkiva: Sure, but that requires some hacks in `Stmt.Import`, `Stmt.Open`. I decide to fix them in later patches.

imkiva: There are two bugs currently: 
- `Data::Option`, it used `import XXX as YYY` then `open YYY`, which results in a wrong cross-link for `YYY` (that's the problem in this comment).
- `SimpleModule`, it opens a local module.

I think the solution to them should be same: by recoding the really-qualified-path in `Stmt.Open`

imkiva:
I guess we need an overhaul of our module names. We now have different types of `ModuleName`, each has a different interpretation. They should be unified in ways like:

- distinguish file-level module from the whole module name (so we can unify `fileModule` and `module` introduced in this PR)
- unqualified module paths should only be used in the resolver (so we can unify `ModulePath` and `ModuleName.Qualified`)

imkiva:
> Resolver 🥲
> 
> I'm modifying the resolver in the classes branch too...

Then let's do it after classes is merged, in case of conflicts

imkiva:
I decide to merge this, and open a new issue for the resolver part. @ice1000

ice1000:
> I decide to merge this, and open a new issue for the resolver part. @ice1000 Please check if this PR conflicts with classes

I don't think there is any yet

ice1000:
bors merge

PR #952, opened 2023-04-03
Literate improvement

imkiva:
- fix #950 
- added yaml frontmatter support

PR #967, opened 2023-04-25
Generalizing Literate

HoshinoTented:
This PR generalizes things related to syntax highlight and md parsing. but Syntax Highlighting looks general enough: `HighlightInfo` is too powerful, and `FaithfulPrettier` is language-unrelated, so we can just move those to the new module without change.


base/src/main/java/org/aya/concrete/remark/AyaLiterate.java
```diff
@@ -0,0 +1,116 @@
...
+import org.aya.tyck.Result;
+import org.aya.util.error.SourcePos;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * @author hoshino
+ */
+public interface AyaLiterate {
+  record AyaError(@NotNull MutableValue<AnyVar> def, @Override @NotNull SourcePos sourcePos) implements Literate {
```
HoshinoTented, line 25: This looks unused, should we remove it?

tools-md/src/main/java/org/aya/literate/Literate.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  record Many(@Nullable Style style, @NotNull ImmutableSeq<Literate> children) implements Literate {
+    @Override public @NotNull Doc toDoc() {
+      var child = Doc.cat(this.children().map(Literate::toDoc));
+      return style == null ? child : Doc.styled(style, child);
+    }
+  }
+
+  class CommonCode implements Literate {
```
imkiva, line 58: Why so many code blocks?

imkiva: We just need one code block class. The derived markdown parser should tell the base parser "the interesting languages".
And when the base parser encounters a code block, check if the language is in that list, if so, parse it to the code block instance, otherwise, parse it to `Literate.Raw(Doc.plain(...)``

HoshinoTented: The base parser probably doesn't know how to construct the inline code block for that language, or you are talking about the fenced code block?

imkiva: Yes. It's fenced code block.

ice1000 (approved):
Awaiting fixing @imkiva's comments

tools-md/src/main/java/org/aya/literate/Literate.java
HoshinoTented: Why don't we parse all the code blocks to `Literate.CodeBlock`?

ice1000: I think we probably can

base/src/main/java/org/aya/concrete/remark/AyaLiterate.java
ice1000: Seems you already removed it

tools-md/src/main/java/org/aya/literate/Literate.java
imkiva: This helps markdown-in-markdown if it is not interesting, see https://github.com/aya-prover/aya-dev/pull/952

ice1000: Kiva said she's tackle this herself

Issue #759, opened 2022-12-04
Library Redefinition


HoshinoTented:
Assume that:
* A library `A`, which defines `prim I`
* A library `B`, which depends on library `A` and defines nothing.
* A library `C`, which depends on library `A` and library `B`, also defines nothing.

<details>
<summary>Aya said:</summary>

```plain
Compiling A
  [Info] Resolving source file dependency
    Done in 158ms
  [Tyck] Prim (Prim.aya)
I like these modules :)
    Library loaded in 412ms
Compiling A
  [Info] Resolving source file dependency
    Done in 1ms
  [Tyck] Prim (Prim.aya)
In file Prim.aya:1:5 ->

  1 | prim I
           ^^

Error: Redefinition of primitive `I'

Resolving interrupted due to:
1 error(s), 0 warning(s).
🔨
```

</details>

> See [here](https://github.com/HoshinoTented/aya-issues/tree/31f6d6603ea0f2f928f9928f203095634e19ce84) for the complete information.

tsao-chi:
😨

tsao-chi:
`A` should be only compiled once by the way

HoshinoTented:
> We should compute SCC of dependencies before compiling the library.

The libraries won't depend on each other, so topo sorting may be enough.

ice1000:
> > We should compute SCC of dependencies before compiling the library.
> 
> The libraries won't depend on each other

Are you sure? What if they declare so?

imkiva:
> > We should compute SCC of dependencies before compiling the library.
> 
> The libraries won't depend on each other, so topo sorting may be enough.

But this is indeed caused by library dependency. And topo sorting is surely a method that computes SCC.

imkiva:
Just relax, I am taking care of this bug.

HoshinoTented:
> And topo sorting is surely a method that computes SCC.

That was amazing! I got it!

ice1000:
Looks like it's closed by #978

PR #978, opened 2023-05-21
Libary Dependency Graph

HoshinoTented:
I hope these utils are helpful in solving #759.

TODO: do we really need those tests?

### DIFF
--- a/cli-impl/src/main/java/org/aya/cli/library/source/LibraryOwner.java
+++ b/cli-impl/src/main/java/org/aya/cli/library/source/LibraryOwner.java
@@ -61,7 +61,13 @@ public interface LibraryOwner {
     }).getOrNull();
   }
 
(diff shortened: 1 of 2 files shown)


HoshinoTented:
These utils work incorrectly with `DiskLibraryOwner`, cause the same library can have different `LibraryOwner` which are not `equals`.

How to check whether two LibraryOwner are equal? `LibraryOwner#underlyingLibrary().libraryRoot().equals`?

ice1000:
Maybe we should create a dedicated field for that purpose...

imkiva:
See the api note on LibraryOwner.

> * @apiNote DO NOT USE this class as Map key, use {@link #underlyingLibrary()} instead.

HoshinoTented:
😱

ice1000 (approved):
bors r+

PR #979, opened 2023-05-27
Let Term

HoshinoTented:
+ [x] `Let Term` and normalization
+ [x] `Nested` interface for nested structure
+ [x] `LetPrettier` for prettying let-like things

TODO:
+ [ ] rendering the type of the bind during rendering `LetTerm`. I may use Synthesizer, but where does the `TyckState` come from?

### DIFF
--- a/base/src/main/java/org/aya/concrete/Expr.java
+++ b/base/src/main/java/org/aya/concrete/Expr.java
@@ -14,6 +14,7 @@ import org.aya.concrete.stmt.QualifiedID;
 import org.aya.concrete.stmt.Stmt;
 import org.aya.concrete.stmt.UseHide;
 import org.aya.generic.AyaDocile;
+import org.aya.generic.Nested;
 import org.aya.generic.ParamLike;
 import org.aya.generic.SortKind;
 import org.aya.guest0x0.cubical.Restr;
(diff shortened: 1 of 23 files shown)


base/src/main/java/org/aya/concrete/Expr.java
```diff
@@ -44,6 +44,43 @@
...
+     *
+     * @return null if the body is not {@link This}
+     * @implSpec {@code tryNested == null || tryNested == body}
+     */
+    @SuppressWarnings("unchecked")
+    default @Nullable This tryNested() {
+      var body = body();
+      var clazz = getClass();
+
+      return clazz.isInstance(body) ? (This) body : null;
```
ice1000, line 80: clazz.cast maybe

ice1000 (commented):
Very nice! I'll go make some minor edits and merge this.

Gj!

base/src/main/java/org/aya/concrete/Expr.java
HoshinoTented: `(This) clazz.cast(body)` ?

HoshinoTented:
😢

HoshinoTented:
What's wrong? Consider we want to `term[x |-> y]` with defEq `z |-> x`, the substituted term still contains `x` after normalization.

So I suspend this PR.

HoshinoTented:
This feature may cost more than its benefit...

PR #982, opened 2023-05-30
Fix baka let and add some improvement

HoshinoTented:
Goals:
+ [x] Nested interface for nested structure (copied from #979)
+ [x] Better prettier for lambda
+ [x] Fixes baka let which has redundant lambda

TODOs:
+ [ ] ~~Better prettier for Pi (if Goal#2 is good)~~ Pi is special

### DIFF
--- a/base/src/main/java/org/aya/generic/Nested.java
+++ b/base/src/main/java/org/aya/generic/Nested.java
@@ -1,0 +1,64 @@
+// Copyright (c) 2020-2023 Tesla (Yinsen) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.generic;
+
+import kala.collection.immutable.ImmutableSeq;
+import kala.collection.mutable.MutableList;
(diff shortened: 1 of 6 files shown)


base/src/main/java/org/aya/generic/Nested.java
```diff
@@ -0,0 +1,64 @@
...
+    This nestedBody = nested;
+    Term body = (Term) nested;
+
+    while (nestedBody != null) {
+      telescope.append(nestedBody.param());
+      body = nestedBody.body();
+      nestedBody = nestedBody.tryNested();
+    }
+
+    return kala.tuple.Tuple.of(telescope.toImmutableSeq(), body);
```
ice1000, line 62: Import

ice1000 (approved):

base/src/main/java/org/aya/generic/Nested.java
HoshinoTented: 😱

PR #992, opened 2023-08-09
Respect `--pretty-dir` for libraries

imkiva:
In the past, the literated library is always saved to `<project-root>/build/pretty`. Now we can specify the output location through `--pretty-dir`


imkiva:
bors merge

PR #996, opened 2023-09-25
Version catalog

ice1000:
Hopefully we can eliminate `deps`
/cc @imkiva need your help with project.version

PR #997, opened 2023-09-26
Cherry pick from #996

imkiva:
image


imkiva:
bors merge

imkiva:
What about moving this workflow to other places like `project-generator`?

ice1000:
> What about moving this workflow to other places like `project-generator`? @ice1000 

I'll see what I can do. You're welcome to do this if you would like to😆

PR #1019, opened 2023-11-23
Some doc related changes

ice1000:
As title

### DIFF
--- a/pretty/src/main/resources/aya-html/highlight-occurrences.js
+++ b/pretty/src/main/resources/aya-html/highlight-occurrences.js
@@ -3,4 +3,4 @@
  * Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
  */
 
-setupHighlight(document);
+highlightFn();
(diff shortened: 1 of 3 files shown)


pretty/src/main/resources/aya-html/highlight-occurrences.js
```diff
@@ -3,4 +3,4 @@
  * Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
  */
 
-setupHighlight(document);
```
imkiva, line 6: I think there are other usages.

imkiva: Maybe in the popup tooltip.

imkiva: show-tooltip-fn.js

imkiva: Use Ctrl+Shift+F please...

PR #1023, opened 2023-12-13
Use Java 21 features

ice1000:
Yeeassss


gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 # The Version of this project, aka, The Aya Theorem Prover.
 # Remove "-SNAPSHOT" suffix and run gradle publish to release a new version.
 # After that, increase the version number and add "-SNAPSHOT" suffix back for next cycle.
-project = "0.30"
```
imkiva, line 5: Please add `-SNAPSHOT` to the version because it is in the main branch, and such a "release" version will prevent the CI from uploading nightly builds to `https://s01.oss.sonatype.org/content/repositories/snapshots`

ice1000: 😨 oh no

PR #1026, opened 2024-02-21
pretty: support ansi16

PhotonQuantum:
This pull request aims to add a new render target, `ANSI16`, to provide better Discord code block support. It's basically Unix format but without 256-color ANSI codes.

Also introduced a quick fix for CLI not respecting output target when using together with the `--fake-literate` option.


ice1000 (approved):
bors r+

ice1000:
Thank you!

PR #1027, opened 2024-03-17
Remove the uses of unnamed parameter (JDK-8313323)

HoshinoTented:
JDK-8313323


HoshinoTented:
This bug seems fixed in 21.0.1.

PR #1047, opened 2024-05-28
Upgrade many things in deps and ci

ice1000:
Also readme


ice1000:
Need to wait for maven central to put upstream 0.24 in it

ice1000:
Can be checked here: https://repo1.maven.org/maven2/org/aya-prover/upstream/build-util/

PR #1074, opened 2024-06-01
Refactor of jit-compiler

HoshinoTented:
This PR removes some useless abstraction of jit-compiler and makes `AbstractSerializer` smaller.


ice1000 (dismissed):
But test fails

PR #1106, opened 2024-06-09
Initial `class` related code

ice1000:
My proposed design on the concrete:

1. `ClazzDecl` only has a telescope, it will never have `result`
2. The telescope's local bindings are copied into a list of `MemberVar`s
3. The resolver should replace the local vars with `MemberVar`s if there exists a corresponding one
4. The tycker should deal with the `self` invocation


ice1000:
I expect this to drastically reduce coverage, in the next PR I'll try to add more tests to save some

PR #1127, opened 2024-06-30
Show Information of Definition in Repl

HoshinoTented:
How to pretty print a core decl?


ice1000:
> How to pretty print a core decl?

I think we need to somehow write a conversion function

PR #1135, opened 2024-10-22
Refactor Module

HoshinoTented:
This PR redesigned the module system:
* ~~`import data::nat` can be referred by `nat` rather than `data::nat`~~
* ambiguous on exported symbol is not allowed, therefore we don't need to solve ambiguity when importing ~~, so `UseHide.Name` should store a single string rather than `QualifiedID`~~
* better candidate implementation (a dedicated class rather than a list/map)


ice1000:
Do you intend to supersede `ModuleExport` with `ModuleExport2`? (and also the other similarly named classes)

HoshinoTented:
> Do you intend to supersede `ModuleExport` with `ModuleExport2`? (and also the other similarly named classes)

Yes, I keep it because I am too lazy to delete it.

HoshinoTented:
The code compiles. However, there are some problems: We have a lot of usage of `import xxx::base`, and they cause name conflicts since the name `base`.

HoshinoTented:
Well, a lot of code is reverted cause:
* We need `ModuleName.Qualified` as the name of module (see last comment)
* I was trying to keep the tree structure of modules, but I found that we can rebuild it from flattened module names, except we lost the layer information (we can not tell whether a module is directly or indirectly importing, but `ResolveInfo` can tell)

HoshinoTented:
However, I found an unnatural place in `CompiledModule::SerExport`, and I fixed it in this PR.

base/src/main/java/module-info.java
```diff
@@ -2,6 +2,7 @@
   requires transitive aya.syntax;
 
   requires static org.jetbrains.annotations;
+  requires jdk.compiler;
```
ice1000, line 5: Marked as TODO

PR #1137, opened 2024-11-13
Reviewing Code, More Documents and Make Notes for Module System

HoshinoTented:


base/src/main/java/org/aya/resolve/context/Candidate.java
```diff
@@ -28,7 +34,7 @@ public sealed interface Candidate<T> {
   /**
    * Returns the only symbol in this candidate, should check {@link #isEmpty()} and {@link #isAmbiguous()} first.
    *
-   * @return the only symbol in this candidate, exception if this candidate {@link #isEmpty()} or {@link #isAmbiguous()}
+   * @return the only symbol in this candidate, exception <b>MAY</b> be thrown if this candidate {@link #isEmpty()} or {@link #isAmbiguous()}
```
ice1000, line 37: Why maybe?

base/src/main/java/org/aya/normalize/Normalizer.java
```diff
@@ -51,6 +51,8 @@ public final class Normalizer implements UnaryOperator<Term> {
       case StableWHNF _, FreeTerm _ -> postTerm;
       case BetaRedex app -> {
         var result = app.make();
+        // TODO: should we use defaultValue? note that this infect the [MemberCall] case in [TermComparator#doCompareUntyped]
```
ice1000, line 54: I think we should use `defaultValue`. When fully normalizing (in which case `defaultValue = postTerm`), we expect everything to be normalized, which includes the arguments.

base/src/main/java/org/aya/normalize/Normalizer.java
```diff
@@ -75,13 +77,15 @@ case FnCall(var fn, int ulift, var args) -> switch (fn) {
...
       case RuleReducer reduceRule -> {
+        // TODO: similar to the case of BetaRedex
         var result = reduceRule.rule().apply(reduceRule.args());
         if (result != null) yield apply(result);
         // We can't handle it, try to delegate to FnCall
         yield reduceRule instanceof RuleReducer.Fn fnRule
           ? apply(fnRule.toFnCall())
           : reduceRule;
       }
+      // TODO: ths case seems unreachable cause there is a fast path at the beginning of this method
```
ice1000, line 88: The fast path works on `term` while this case works on `postTerm`

base/src/main/java/org/aya/resolve/context/Candidate.java
HoshinoTented: `Imported#get` won't throw if there are many candidates

base/src/main/java/org/aya/normalize/Normalizer.java
HoshinoTented: ^ I am not sure

ice1000:
CI fails

HoshinoTented:
Seems we didn't check `isAmbiguous` somewhere.

base/src/main/java/org/aya/normalize/Normalizer.java
ice1000: `someCastTerm` should not unfold to `innerestTerm`

ice1000: If it does then the code is not correct IMO

HoshinoTented: `someCastTerm` doesn't unfold to `innerestTerm` but `someCastTerm .field` unfolds to `innerestTerm .field`. In fact,
`.field` is the following function:
```aya
.field (new impls) = ...
.field (cast inner remember forget) = if field in remember or forget then ... else inner .field
```

ice1000: I see I see I think this reduction is ok.

PR #1147, opened 2024-11-18
Fix two issues related to arrays and meta lit terms, implement `debug-show-shapes`

ice1000:
As title


ice1000:
-0.03 is good enough

PR #1179, opened 2024-12-04
More stdlib

HoshinoTented:
Add more stuff to standard library and find some bugs

### DIFF
--- a/cli-impl/src/test/resources/shared/src/data/either/base.aya
+++ b/cli-impl/src/test/resources/shared/src/data/either/base.aya
@@ -1,0 +1,25 @@
+open import relation::binary::path using (= , refl)
+
+open inductive Either (A B : Type) | left A | right B
+
+variable A B : Type
+
+def Either-rec
+  (P : Type)
+  (lrec : A -> P)
+  (rrec : B -> P)
+  (e : Either A B)
+  : P => Either-ind e (fn _e => P) (fn a _p => lrec a) (fn b _p => rrec b)
+
+def Either-ind
+  (e : Either A B)
+  (P : Either A B -> Type)
+  (lrec : Fn (a : A) -> e = left a -> P (left a))
+  (rrec : Fn (b : B) -> e = right b -> P (right b))
+  : P e elim e
+| left a => lrec a refl
+| right b => rrec b refl
+
+def swap (Either A B) : Either B A
+| left a => right a
+| right b => left b
--- a/cli-impl/src/test/resources/shared/src/data/either/base.aya
+++ b/cli-impl/src/test/resources/shared/src/data/either/base.aya
@@ -1,25 +1,25 @@
 open import relation::binary::path using (= , refl)
 
-open inductive Either (A B : Type) | left A | right B
+open inductive Sum (A B : Type) | inl A | inr B
 
 variable A B : Type
 
-def Either-rec
+def Sum-rec
+  (e : Sum A B)
   (P : Type)
   (lrec : A -> P)
   (rrec : B -> P)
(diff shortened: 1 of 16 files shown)


HoshinoTented:
```aya
open import data::either
open import relation::nullary::empty
open import relation::unary::negation

def what (A : Type) (lem : Either A (neg A)) (a : A) : A =>
  Either-rec _
  (fn a => a)
  (fn na => exfalso (na a))
  lem
```

<details>
<summary>Stacktrace</summary>

```plain
Exception in thread "main" java.lang.AssertionError: Failed with `class org.aya.tyck.error.BadTypeError`: In file bad.aya:8:21 ->

  6 |     Either-rec _
  7 |     (fn a => a)
  8 |     (fn na => exfalso (na a))
    |                        ^--^

Error: Unable to apply the expression
         na a
       because the type of what you applied is not a Pi/Path type, but instead:
         A -> Empty

at (211-214) [8,21-8,24]
```

</details>

cli-impl/src/test/resources/shared/src/data/either/base.aya
```diff
@@ -0,0 +1,25 @@
...
+  (lrec : A -> P)
+  (rrec : B -> P)
+  (e : Either A B)
+  : P => Either-ind e (fn _e => P) (fn a _p => lrec a) (fn b _p => rrec b)
+
+def Either-ind
+  (e : Either A B)
+  (P : Either A B -> Type)
+  (lrec : Fn (a : A) -> e = left a -> P (left a))
+  (rrec : Fn (b : B) -> e = right b -> P (right b))
```
ice1000, line 18: This is probably nonstandard, usually e is the last parameter, and you don't need these e = left a parameters

cli-impl/src/test/resources/shared/src/data/either/base.aya
```diff
@@ -0,0 +1,25 @@
+open import relation::binary::path using (= , refl)
+
+open inductive Either (A B : Type) | left A | right B
```
ice1000, line 3: Can you rename Either into Sum and left to inl, right to inr?

cli-impl/src/test/resources/shared/src/data/list/properties.aya
```diff
@@ -19,3 +19,32 @@ def map-id (l : List A) : map (\x => x) l = l
...
+
+def ++-assoc (xs ys zs : List A) : (xs ++ ys) ++ zs = xs ++ (ys ++ zs) elim xs
+| [ ] => refl
+| x :< xs => pmap (x :<) (++-assoc xs ys zs)
+
+private def rev'-map (f : A -> B) (buf xs : List A) : map f (rev' buf xs) = rev' (map f buf) (map f xs) elim xs
+| [ ] => refl
+| x :< xs => rev'-map f (x :< buf) xs
+
+def rev-commu-map (f : A -> B) (xs : List A) : map f (rev xs) = rev (map f xs) => rev'-map f [ ] xs
```
ice1000, line 31: I think you can call it rev-map

ice1000 (commented):
Rest lgtm

cli-impl/src/test/resources/shared/src/data/either/base.aya
HoshinoTented: In case the user wants to rewrite something

ice1000: `lrec` technically already assumes `e = left a` IMO

ice1000:
Any idea why CI fails?

PR #1197, opened 2024-12-08
Free Java

HoshinoTented:
This PR proposes a `FreeJavaBuilder` interface that allows users to construct the FreeJava in Java source code.

## Free Java

FreeJava is the AST of Java source code, which can be compiled/interpreted to different targets, such as bytecode, source code, or 
 itself.

## FreeJavaBuilder

`FreeJavaBuilder` is an interface that allows users to construct a FreeJava in builder form. Those methods are just like the constructor of FreeJava, but they will be interpreted immediately.

## Goal

`FreeJavaBuilder` can be used to construct FreeJava, then we can perform optimization, and translate to another target.

## TODO

- [x] Complete FreeJavaBuilder
- [x] Interpret to source code
- [ ] Interpret to bytecode
- [ ] Interpret to free java and optimization

## Note

This PR is supposed to run on JDK 24 which version with stable Classfile API.


ice1000:
Can we decouple free Java into a separate subproject?

jit-compiler/src/main/java/org/aya/compiler/free/FreeUtils.java
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2020-2024 Tesla (Yinsen) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.compiler.free;
+
+import org.jetbrains.annotations.NotNull;
+
+import java.lang.constant.ClassDesc;
+
+public final class FreeUtils {
```
ice1000, line 9: Can you rename it to `FreeUtil`?

ice1000:
Do we essentially need `FreeJavaBuilder`? Can't we just `new FreeJava` objects?

mio-19:
> FreeJava is the AST of Java source code, which can be compiled/interpreted to different targets, such as bytecode, source code, or
> itself.

What does it mean for FreeJava to be compiled or interpreted to itself?

ice1000:
> > FreeJava is the AST of Java source code, which can be compiled/interpreted to different targets, such as bytecode, source code, or
> > itself.
> 
> What does it mean for FreeJava to be compiled or interpreted to itself?

Just that the identity function compiles free java to free java. I think it doesn't matter much

HoshinoTented:
> > > FreeJava is the AST of Java source code, which can be compiled/interpreted to different targets, such as bytecode, source code, or
> > > itself.
> > 
> > 
> > What does it mean for FreeJava to be compiled or interpreted to itself?
> 
> Just that the identity function compiles free java to free java. I think it doesn't matter much

This progress corresponding to optimization.

jit-compiler/src/main/java/org/aya/compiler/ModuleSerializer.java
```diff
@@ -64,7 +65,7 @@ private void doSerialize(ModuleResult unit) {
       IterableUtil.forEach(unit.defs, this::appendLine, this::doSerialize));
   }
 
-  @Override public ModuleSerializer serialize(ModuleResult unit) {
+  @Override public ModuleSerializer serialize(@NotNull FreeCodeBuilder builder, ModuleResult unit) {
```
mio-19, line 68: I noticed that the @NotNull annotation is violated in FileSerializer. Was this intentional?

ice1000: The same violation pattern happens almost everywhere. I think it's just because the code is WIP

ice1000:
Is it the compilation for `match` expressions TODO?

ice1000:
The tests are indeed failing bc of the TODO, I thikn

ice1000:
This should do

PR #1213, opened 2024-12-13
Improve the safety of `compareApprox`

ice1000:
Do not early-solve metas in compareApprox

### DIFF
--- a/base/src/main/java/org/aya/tyck/TyckState.java
+++ b/base/src/main/java/org/aya/tyck/TyckState.java
@@ -85,7 +85,8 @@ public final class TyckState {
   @ApiStatus.Internal
   public void solve(MetaVar meta, Term candidate) { solutions.put(meta, candidate); }
 
-  private boolean solveEqn(@NotNull Reporter reporter, @NotNull Eqn eqn, boolean allowDelay) {
+  @ApiStatus.Internal
+  public boolean solveEqn(@NotNull Reporter reporter, @NotNull Eqn eqn, boolean allowDelay) {
     var unifier = new Unifier(this, eqn.localCtx, reporter, eqn.pos, eqn.cmp, allowDelay);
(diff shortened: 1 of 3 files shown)


ice1000:
The program below is considered:
```aya
open inductive Nat | zero | succ Nat
open inductive Store (n : Nat) | unit
def make (n : Nat) : Store n => unit

def nonInjective Nat : Nat | 0 => 0 | _ => 0
def injective Nat : Nat | x => x
def the (A : Type) (x : A) : A => x

def test => let
  | x : Nat := _
  | y := the (Store (nonInjective 1)) (make (nonInjective x))
  in the (Store (injective 2)) (make (injective x))
```
This will fail to type check, but if we assign `x : Nat := 2` then it will. This is a case where lossy-unification produces the wrong result.

In this PR, I solved a special case of this: we might refute `x := 1` if `nonInjective` has another parameter that causes the unification to fail, and this will cause `x` to be solved differently.

ice1000:
A slight simplification of the code:

```
open inductive Nat | zero | succ Nat
open inductive Store (n : Nat) | make

def nonInjective Nat : Nat | 0 => 0 | _ => 0
def the (A : Type) (x : A) : A => x

def test => let
  | x : Nat := _
  | y := the (Store (nonInjective 1)) (make {nonInjective x})
  in the (Store (the _ 2)) (make {the _ x})
```

ice1000:
In this PR we improve the unifier so this new code can be accepted:

```aya
open inductive Nat | zero | succ Nat
open inductive Store (n : Nat) | make

def nonInjective Nat Nat : Nat => 0
def the (A : Type) (x : A) : A => x

def test => let
  | x : Nat := _
  | y := the (Store (nonInjective 1 114)) (make {nonInjective x 514})
  in the (Store (the _ 2)) (make {the _ x})
```

While the previous Aya rejects it

PR #1224, opened 2024-12-16
Reimplement match as a call

### LINKED ISSUE #1219: Implement match term elimination as an untyped function definition
Something with an `invoke` on it would be great, I think

ice1000:
Should resolve #1219, but we will have to JIT compile this kind of match


ice1000:
Merge now, because 6 greens is really rare

Issue #1220, opened 2024-12-15
Support dependent types in `variable`


ice1000:
Right now, `variable A : B` cannot have inter-dependencies.

Would be nice to support

```aya
variable A : Type
variable a : A
```

and using `a` automatically introduces `A`.

PR #1226, opened 2024-12-17
Support dependent types in `variable`

mio-19:
fix https://github.com/aya-prover/aya-dev/issues/1220

passed:
```aya
variable A : Type
variable a : A
```

### DIFF
--- a/base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
+++ b/base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
@@ -139,7 +139,6 @@ public record ExprResolver(
               // Ordered set semantics. Do not expect too many generalized vars.
               var owner = generalized.owner;
               assert owner != null : "Sanity check";
-              introduceDependencies(generalized);
               var param = owner.toExpr(false, generalized.toLocal());
               allowedGeneralizes.put(generalized, param);
               addReference(owner);
--- a/base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
+++ b/base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
@@ -280,10 +280,13 @@ public record ExprResolver(
   public @NotNull AnyVar resolve(@NotNull QualifiedID name) {
     var result = ctx.get(name);
     if (result instanceof GeneralizedVar gvar) {
-      // Ensure all dependencies are introduced
-      introduceDependencies(gvar);
       var gened = allowedGeneralizes.getOrNull(gvar);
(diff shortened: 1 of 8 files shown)


base/src/main/java/org/aya/resolve/visitor/ExprResolver.java
```diff
@@ -142,6 +142,7 @@ public ExprResolver(@NotNull Context ctx, boolean allowGeneralizing) {
               // Ordered set semantics. Do not expect too many generalized vars.
               var owner = generalized.owner;
               assert owner != null : "Sanity check";
+              introduceDependencies(generalized);
```
ice1000, line 145: Are you sure you need it here? It seems you called it twice

ice1000:
Can you write a test?

ice1000:
Why are there 39 commits in the PR, and 1000+ loc changes?

image

mio-19:
> Why are there 39 commits in the PR, and 1000+ loc changes?
> 
> image

I entered a wrong git rebase command

HoshinoTented:
Perhaps we need a test for the use of those variables.

syntax/src/main/java/org/aya/syntax/ref/GeneralizedVar.java
```diff
@@ -6,17 +6,27 @@
 import org.aya.util.error.SourceNode;
 import org.aya.util.error.SourcePos;
 import org.jetbrains.annotations.NotNull;
+import kala.collection.immutable.ImmutableSeq;
 
 public final class GeneralizedVar implements AnyVar, SourceNode {
   public final @NotNull String name;
   public final @NotNull SourcePos sourcePos;
   public Generalize owner;
+  private @NotNull ImmutableSeq<GeneralizedVar> dependencies = ImmutableSeq.empty();
```
HoshinoTented, line 15: A `@Nullable` dependencies with a `null` initial value can be better because we shouldn't use them before resolving.

ice1000:
When I wrote some tests I realized this implementation doesn't work at all. Writing down a definition and use a `variable` will raise scope error, which is caused by incoherent generation of `LocalVar`.

Repro:

```
prim I prim Path
variable A : Type
def infix = (a b : A) => Path (fn i => A) a b

variable a b : A
def test (a = a) => a
```

I don't think this is going to be fixed, so I'm closing this and I'm gonna implement this myself

HoshinoTented:
😭

ice1000:
😭

mio-19:
😿

PR #1233, opened 2024-12-18
Release 0.36, upgrade to Java 22

ice1000:
I need to go home to actually publish it


ice1000:
We need to skip Java 23 bc of https://bugs.openjdk.org/browse/JDK-8341901

PR #1235, opened 2024-12-19
Fix Wrong Implementation of `JitCon#selfTele`

HoshinoTented:
The old `JitCon#selfTele` returns with lost var, which is incorrect. This PR intends to fix it.
However, the test didn't fail due to incorrect behavior, which may imply potential bugs.


ice1000:
Your the new STLC code might be a good test case

PR #1247, opened 2024-12-21
Report Error When Using Implicit Patterns with Eliminator

HoshinoTented:


ice1000:
Why so many tests fail?

PR #1280, opened 2025-01-06
Fix #1274

### LINKED ISSUE #1274: Optimize simple calls to ImmSeq.from

HoshinoTented:
Fix #1274


HoshinoTented:
The `@see` sequence is used for resolving the factory functions by IDE, in case kala's breaking changes (although it is almost impossible).

PR #1281, opened 2025-01-07
Pretty Print and Highlight for Pragma

HoshinoTented:


producer/src/main/java/org/aya/producer/AyaProducer.java
```diff
@@ -293,19 +297,31 @@ private void pragma(GenericNode<?> node, Decl decl) {
...
+
       switch (name) {
-        case "suppress" -> args.forEach(arg -> {
-          for (var suppress : Suppress.values()) {
-            if (arg.tokenText().contentEquals(suppress.name())) {
-              decl.suppresses.add(suppress);
-              return;
+        case Constants.PRAGMA_SUPPRESS -> {
+          // TODO: use MutableEnumSet
+          MutableSet<WithPos<Suppress>> sups = MutableSet.create();
```
ice1000, line 307: I think I used a set because I want each suppress to be unique, but now with a source pos the same warning can be suppresses many times in the same set.
I suggest using a list instead

HoshinoTented: I am not sure if we really need the source pos.

ice1000: I think we do, it's needed for faithful highlight of them

PR #1282, opened 2025-01-09
Use `bindAllFrom` primitively

### LINKED ISSUE #1236: Instead of implementing `Term::bindAt` primitively, we should implement `bindTele` primitively
https://github.com/aya-prover/aya-dev/blob/70893a5a0ec59c7e2e76c3d9c24db36bbd92b481/syntax/src/main/java/org/aya/syntax/core/term/Term.java#L61-L65

Right now `bindTele` is based on `bindAt`, I think it should be the other way around: `bindAt` should be based on `bindTele`.

HoshinoTented:
Fix #1236


ice1000:
Thank you!

PR #1283, opened 2025-01-16
Irrefutable Lambda

### LINKED ISSUE #94: Irrefutable patterns in lambda
As title

HoshinoTented:
This PR fixes #94 by:
* AyaProducer producing `Expr.IrrefutableLam`
* Desugar `Expr.IrrefutableLam` to `Expr.Lambda`

TODO:
+ [x] producer
+ [x] question: should we resolve `Expr.Lambda` even if it is a desugar output?


ice1000:
> * [ ]  question: should we resolve `Expr.Lambda` even if it is a desugar output?

Desugar happens after resolving so this should be not applicable

HoshinoTented:
How about the interaction with pushing pattern from lambda in pattern matching? i.e.:
```aya
def what (A : Type) (n : Nat)
| A => fn zero => ...
| A => fn suc n => ...
```
becomes
```aya
def what (A : Type) (n : Nat)
| A, zero => ...
| A, suc n => ...
```

ice1000:
> How about the interaction with pushing pattern from lambda in pattern matching? i.e.:
> 
> ```
> def what (A : Type) (n : Nat)
> | A => fn zero => ...
> | A => fn suc n => ...
> ```
> 
> becomes
> 
> ```
> def what (A : Type) (n : Nat)
> | A, zero => ...
> | A, suc n => ...
> ```
> 
> This may require the desugar in tyck rather than just after resolving.

🤯🤯🤯.... I think it's ok to not implement this feature. It's unreasonably complicated. Pushing lambda was meant to be a simple feature

PR #1291, opened 2025-01-28
remove a task which depends on non exisiting :lsp:jlink

mio-19:
This is causing a problem on a certain ide setup

```

[error] FAILURE: Build failed with an exception.

* What went wrong:
Could not create task ':githubActions'.
> A dependency must not be empty

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

CONFIGURE FAILED in 222ms
```


ice1000:
It should be changed to `:ide-lsp:jlink`, but we no longer use it, I think

PR #1296, opened 2025-01-31
Reimplement dependency in variables

ice1000:
Supersedes #1226, working on tests rn


ice1000:
This is more complicated than I thought

PR #1297, opened 2025-01-31
Classfile Morphism

HoshinoTented:
This PR introduces a morphism that targets jvm bytecode, which is much more efficient than java source code. However, jvm bytecode is so low-level and will have a lot of potential problems. We need more tests on this.

I also need more advice about bytecode generation.

### DIFF
--- a/jit-compiler/src/main/java/org/aya/compiler/free/morphism/asm/AsmClassBuilder.java
+++ b/jit-compiler/src/main/java/org/aya/compiler/free/morphism/asm/AsmClassBuilder.java
@@ -4,9 +4,6 @@ package org.aya.compiler.free.morphism.asm;
 
 import java.lang.constant.*;
 import java.lang.invoke.LambdaMetafactory;
-import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
-import java.lang.invoke.MethodType;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -38,7 +35,7 @@ public final class AsmClassBuilder implements FreeClassBuilder {
   public final @NotNull MutableMap<FieldRef, Function<FreeExprBuilder, FreeJavaExpr>> fieldInitializers = MutableLinkedHashMap.of();
   private int lambdaCounter = 0;
 
-  /// @see java.lang.invoke.LambdaMetafactory#metafactory(MethodHandles.Lookup, String, MethodType, MethodType, MethodHandle, MethodType)
+  /// @see java.lang.invoke.LambdaMetafactory#metafactory
   private final @NotNull LazyValue<MethodHandleEntry> lambdaBoostrapMethodHandle;
 
   public AsmClassBuilder(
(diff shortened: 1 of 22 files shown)


jit-compiler/src/main/java/org/aya/compiler/free/morphism/asm/AsmUtil.java
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2020-2025 Tesla (Yinsen) Zhang.
+// Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
+package org.aya.compiler.free.morphism.asm;
+
+public final class AsmUtil {
+  private AsmUtil() { }
+
+}
```
ice1000, line 8: Still needed?

ice1000 (commented):
Haven't done reading everything yet

jit-compiler/src/main/java/org/aya/compiler/free/morphism/asm/AsmClassBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+import org.glavo.classfile.ClassBuilder;
+import org.glavo.classfile.attribute.NestMembersAttribute;
+import org.glavo.classfile.constantpool.InvokeDynamicEntry;
+import org.glavo.classfile.constantpool.MethodHandleEntry;
+import org.jetbrains.annotations.NotNull;
+
+public final class AsmClassBuilder implements FreeClassBuilder {
+  public final @NotNull ClassDesc owner;
+  public final @NotNull ClassDesc ownerSuper;
+  public final @NotNull ClassBuilder writer;     // I am sorry
```
ice1000, line 35: Why?

jit-compiler/src/main/java/org/aya/compiler/free/morphism/asm/AsmUtil.java
HoshinoTented: may not

jit-compiler/src/main/java/org/aya/compiler/free/morphism/asm/AsmClassBuilder.java
HoshinoTented: i don't know

jit-compiler/src/main/java/org/aya/compiler/free/morphism/asm/AsmClassBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+      lambdaMethodHandle,
+      interfaceMethodDesc   // just hope it doesn't cause any trouble
+    ).map(pool::loadableConstantEntry).asJava());
+
+    return pool.invokeDynamicEntry(bsmEntry, nameAndType);
+  }
+
+  public void postBuild() {
+    if (nestedMembers.isNotEmpty()) {
+      writer.with(NestMembersAttribute.of(nestedMembers.map(x -> writer.constantPool().classEntry(x)).asJava()));
```
ice1000, line 166: Method reference?

Issue #1302, opened 2025-02-02
Avoid generating `MutableSeq` in `invoke`s


ice1000:
In the JIT-compiled defs.

Right now we generate these `MutableSeq` to store some local variables of type `Term`:
https://gist.github.com/ice1000/70fd92ce92bd744c663e4f7e81d61192#file-_baka-java-L529

But they can be replaced with stack variables.

They were created to hack javac, who insist on "captured variables must be final or effectively final", while these local vars are re-assigned in the pattern matching code, so javac rejects them. But this is not a limitation of the JVM itself, and since we're generating bytecode directly, we should flatten these `MutableSeq` onto the stack.

PR #1303, opened 2025-02-03
Avoid using `MutableSeq` for local term variables #1302

linxuanm:
Addresses #1302.


jit-compiler/src/main/java/org/aya/compiler/serializers/MatchySerializer.java
```diff
@@ -72,9 +72,10 @@ private void buildInvoke(
 
     var matching = unit.clauses().map(clause ->
       new PatternSerializer.Matching(clause.bindCount(), clause.patterns(),
-        (ps, cb, bindCount) -> {
-          var resultSeq = AbstractExprializer.fromSeq(cb, Constants.CD_Term, ps.result.ref(), bindCount);
+        (ps, cb, _) -> {
+          var resultSeq = ps.result.map(LocalVariable::ref);
```
HoshinoTented, line 76: I am not sure if all `result` are available (not null).

linxuanm: Sorry I don't think I follow. I think this should be fine as long as `ps.result` isn't null? `result` should be a sequence of non-null `LocalVariable` populated in `PatternSerializer#serialize`.

HoshinoTented: All `LocalVariable` is always not null, but not the value they represent at runtime. The value of some `result` may be null and cannot be used for `serializeTermUnderTele`

HoshinoTented: I mean we should use the 3rd parameter.

linxuanm: Ah I see.

HoshinoTented (dismissed):
LGTM

ice1000 (dismissed):

Issue #1301, opened 2025-02-02
Use `indexedLines` to cache the lineCol-offset correspondence


ice1000:
Instead of using this shit

![Image](https://github.com/user-attachments/assets/a80c4cc2-01ea-4f4d-9407-539fcbda6c4a)

ice1000:
https://github.com/aya-prover/aya-dev/blob/7d1ffdcaff14cd3b804ef4b972348ae9ced01e43/tools-kala/src/main/java/org/aya/util/more/StringUtil.java#L31

ice1000:
Need to refactor that, ofc

linxuanm:
#1307

ice1000:
> #1307 

No need to send a reply. GitHub supports linking a branch/PR with an issue

PR #1307, opened 2025-02-07
Fast LineColumn conversion (for #1301)

linxuanm:
Caches line positions in `SourceFile`.

### DIFF
--- a/tools/src/main/java/org/aya/util/StringUtil.java
+++ b/tools/src/main/java/org/aya/util/StringUtil.java
@@ -2,9 +2,9 @@
 // Use of this source code is governed by the MIT license that can be found in the LICENSE.md file.
 package org.aya.util;
 
-import kala.collection.immutable.ImmutableSeq;
 import kala.collection.immutable.primitive.ImmutableIntArray;
 import kala.collection.mutable.primitive.MutableIntList;
+import kala.value.primitive.MutableIntValue;
 import org.jetbrains.annotations.NotNull;
 
 import java.time.ZonedDateTime;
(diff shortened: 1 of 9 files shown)


tools/src/main/java/org/aya/util/StringUtil.java
```diff
@@ -26,18 +25,15 @@ public interface StringUtil {
...
-    var lines = ImmutableSeq.from(str.lines());
-    var results = MutableList.<IntObjTuple2<String>>create();
-
+  static @NotNull ImmutableSeq<Integer> indexedLines(@NotNull String str) {
+    var results = MutableList.<Integer>create();
     var index = 0;
-    for (var line : lines) {
-      results.append(IntObjTuple2.of(index, line));
-      index = index + line.length() + 1;    // 1 for the line separator
+    for (var line: ImmutableSeq.from(str.lines())) {
```
HoshinoTented, line 33: The construction of `ImmutableSeq` is used for iteration only, which is quite heavy. I think we can:
* use `str.lines().forEach(...)`, but we need to replace `int index` with `MutableIntValue index`.
* use `str.lines().iterator()` with a `while` loop, which is not very pretty.

tools/src/main/java/org/aya/util/error/SourceFile.java
```diff
@@ -18,8 +20,10 @@
 public record SourceFile(
   @NotNull String display,
   @NotNull Option<Path> underlying,
-  @NotNull String sourceCode
+  @NotNull String sourceCode,
+  @NotNull ImmutableIntArray lineOffsets
```
ice1000, line 24: Is it possible to store this in the producer instead?

ice1000: Nvm

ice1000 (approved):

PR #1309, opened 2025-02-08
Add some symbol information and misc

HoshinoTented:
Now the expression evaluator in the debugger works fine, but IDEA still cannot perform step over/into precisely, perhaps we need to supply the line number.


ice1000 (approved):
💯

PR #1313, opened 2025-02-11
More lsp changes

HoshinoTented:
TODO:
- [ ] Fix potential threading problem of mutable states in `AyaLanguageServer`

### DIFF
--- a/ide-lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
+++ b/ide-lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
@@ -95,8 +95,8 @@ public class AyaLanguageServer implements LanguageServer {
     return SeqView.narrow(mockLibraries(path));
   }
 
-  /// Check whether the project/the aya file {@param projetOrFile} represents is registered in this {@link AyaLanguageServer}.
-  private @Nullable LibraryOwner getRegisteredLibrary(@NotNull ProjectOrFile projectOrFile) {
+  /// Check whether the project/the aya file {@param projectOrFile} represents is registered in this {@link AyaLanguageServer}.
+  public @Nullable LibraryOwner getRegisteredLibrary(@NotNull ProjectOrFile projectOrFile) {
     return libraries.getOrNull(projectOrFile.path());
   }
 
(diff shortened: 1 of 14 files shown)


ide-lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
```diff
@@ -83,36 +83,48 @@ public AyaLanguageServer(@NotNull CompilerAdvisor advisor, @NotNull AyaLanguageC
...
+      return tryAyaLibrary(project);
+    }
+
+    // resolved is Directory or File
+    return SeqView.narrow(mockLibraries(resolved.path()));
+  }
+
+  /// Check whether the project/the aya file {@param projectOrFile} represents is registered in this {@link AyaLanguageServer}.
+  public @Nullable LibraryOwner getRegisteredLibrary(@NotNull ProjectPath projectOrFile) {
+    return libraries.getOrNull(FileUtil.canonicalize(projectOrFile.path()));
```
ice1000, line 106: Is this path always canonicalized?

ide-lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
```diff
@@ -124,10 +136,12 @@ public SeqView<LibraryOwner> registerLibrary(@NotNull Path path) {
...
 
-  private ImmutableSeq<WsLibrary> mockLibraries(@NotNull Path path) {
-    var mocked = AyaFiles.collectAyaSourceFiles(path, 1).map(WsLibrary::mock);
-    libraries.appendAll(mocked);
-    return mocked;
+  private SeqView<WsLibrary> mockLibraries(@NotNull Path path) {
+    var mocked = AyaFiles.collectAyaSourceFiles(path, 1)
+      .map(f -> Tuple.of(f, WsLibrary.mock(f)));
+
+    mocked.forEach(libraries::put);
```
ice1000, line 143: Can use `onEach`

ide-lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
HoshinoTented: my bad

ide-lsp/src/main/java/org/aya/lsp/server/AyaLanguageServer.java
ice1000: Next PR

ice1000:
I don't see how the TODOs can be fixed immediately, nor am I sure about what should we lock. Naïvely we would just turn all the methods who mutate `libraries()` synchronized, but I don't know if it's correct.

PR #1315, opened 2025-02-11
Use `onEach` and refactor some `stream` to `view`

ice1000:
Kala 0.81


ice1000:
MutableLinkedSet is missing
https://github.com/Glavo/kala-common/blob/c57b5104bbb28530caa1c175ead5778d1b243d2e/CHANGELOG.md?plain=1#L65

PR #1323, opened 2025-02-13
Rename keyword `partial` to `nonterminating`

ice1000:
As title


mio-19:
This keyword is so long

ice1000:
> This keyword is so long

True. It's shorter than the Agda equivalent though.

PR #1326, opened 2025-02-15
Perform WHNF when pattern matching and invoke

HoshinoTented:
Currently, our PatternSerializer missing the part of whnf the argument, which may stuck in certain situations. Also, `TermExperialize` is responsible for doing some `Normalizer` work, such as normalizing the argument before `invoke`.

This PR:
+ [x] Add a `UnaryOperator` argument (can be `Normalizer`) to `JitFn#invoke`, `JitMatchy#invoke`, and `JitCon#isAvailable`
+ [x] Use `InvokeSignatureHelper` to handle the change above.


ice1000 (approved):
Lgtm

PR #1352, opened 2025-04-26
Implementing LSP Completion

HoshinoTented:
This PR:
- [x] Orgalize resolving so that we can provide more resolving information for completion
- [x] Collect the environment at the given position so that we can provide all necessary information for completion.

TODO:
- [x] Refactor error handling of resolving/compiling
- [ ] Provide completion information by the identifier at the given position, for example: provide all symbols in `Foo::Nat` if `Foo::Nat:_`, provide all symbols in `Foo` starts with `Nat` if `Foo::Nat_`.
- [ ] Testing about completion (especially aya-vscode)


ice1000:
Missing:

> * Provide completion information by the identifier at the given position, for example: provide all symbols in `Foo::Nat` if `Foo::Nat:_`, provide all symbols in `Foo` starts with `Nat` if `Foo::Nat_`.
> * [ ]  Testing about completion (especially aya-vscode)

ice1000:
image

Good enough

PR #1354, opened 2025-05-03
PR to PR

ice1000:
Replace pervasive exception with `result` and CPS `Reporter`


ice1000:
Tests passed locally let's go!!!!!!!

PR #1360, opened 2025-06-07
Better Completion and some Test Resource

HoshinoTented:
This PR:
- [x] Provides better completion, such as licit of parameters, hides invisible variables, provides data name of constructors
- [x] Add some interesting proof
- [x] Miscellaneous changes

TODO in future:
- [ ] `Commad.Module` and `Expr.LetOpen` should store the module context they own, then provide more completion items by those contexts.


ide-lsp/src/test/java/org/aya/lsp/CompletionTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Test public void testCompletionProvider() {
+    // TODO: check by human eyes, sorry
+    var client = launch(TEST_LIB);
+    client.execute(compile((_, _) -> { }));
+    var source = client.service.find(TEST_LIB.resolve("src").resolve("HelloWorld.aya"));
+    assert source != null;
+    var xy = new XY(14, 55);        // _c a a
+    // XY: 14, 52
+    var list = CompletionProvider.completion(source, xy, AyaDocile::easyToString).items;
```
ice1000, line 82: Can we assert something?

PR #1391, opened 2025-10-20
Improve IssueSetup

HoshinoTented:
Fix some safety issue and write metadata.json to issue tracker working directory


cli-impl/src/test/java/org/aya/test/cli/IssueTrackerTest.java
```diff
@@ -10,19 +10,53 @@
...
+      inductive Nat | zro | suc Nat
+      
+      def what => 0""", fooAya);
+  }
+
+  @Test
+  public void testDevil() {
+    var base = Path.of("src");
+
+    var file = new IssueParser.File("../../../../../../etc/shadow", "");
```
ice1000: But this file can exist, right?

ice1000: I see

PR #1427, opened 2025-11-16
Instance Filterrrrrrrrrrr

HoshinoTented:
Also there are some dblity mismatch in `TermComparator` and `Unifier`, should be eliminated.

### DIFF
--- a/base/src/main/java/org/aya/normalize/Finalizer.java
+++ b/base/src/main/java/org/aya/normalize/Finalizer.java
@@ -17,8 +17,10 @@ import org.aya.syntax.core.term.call.MatchCall;
 import org.aya.syntax.core.term.call.MetaCall;
 import org.aya.syntax.core.term.repr.MetaLitTerm;
 import org.aya.syntax.ref.MetaVar;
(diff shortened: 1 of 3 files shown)


base/src/main/java/org/aya/states/TyckState.java
```diff
@@ -114,7 +117,36 @@ public void solveMetas(@NotNull Reporter reporter) {
...
+      return switch (it) {
+        // try replace all param of [def] with meta, then solve by unify with [insted.type()]
+        case Instance.Global(var def) -> throw new UnsupportedOperationException("TODO");
+        case Instance.Local(var ref, var ty) -> {
+          // TODO: which unifier we should use?
+          // i guess this won't cause infinite recursion, as the context of `meta` doesn't contain itself
+          Unifier someUnifier = null;
+          var result = someUnifier.compare(insted.type(), ty, null);
+          // still keep unsure
+          // TODO: what if YES for 1 instance and UNSURE for many other instances, is it possible?
```
ice1000: It will be possible, and I think we should keep all the unsure instances in that case.

PR #1449, opened 2025-12-16
Refactor CompiledModule/ResolveInfo

HoshinoTented:
This PR slightly refactors CompiledModule and ResolveInfo, by serializing/deserializing the `ResolveInfo.thisModule.export` directly instead of serializing/deserializing the context-related commands (such as Command.Import).

Q: Why don't serialize context-related commands?
A: Commands are imperative, thus the order is important, furthermore, some definitions act like a command, such as `inductive`, which introduces a sub-module, and it is very stupid to store the order information...

Although serializing `ModuleExport` directly may increase the size of `,ayac`, especially `prelude.aya`.

### DIFF
--- a/jit-compiler/src/main/java/org/aya/compiler/CompiledModule.java
+++ b/jit-compiler/src/main/java/org/aya/compiler/CompiledModule.java
@@ -244,13 +244,13 @@ public record CompiledModule(
     }
 
     /// Load any public definition
-    public @NotNull AnyDefVar load(@NotNull QName name) {
+    public @NotNull CompiledVar load(@NotNull QName name) {
       if (name.module().fileModule().equals(thisModulePath)) {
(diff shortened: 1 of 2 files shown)


jit-compiler/src/main/java/org/aya/compiler/CompiledModule.java
```diff
@@ -159,12 +143,172 @@ record RenameData(boolean reExport, QName name, SerRenamedOp renamed) { }
...
+      };
+    }
+
+    /// Load any public definition
+    public @NotNull AnyDefVar load(@NotNull QName name) {
+      if (name.module().fileModule().equals(thisModulePath)) {
+        return new CompiledVar(this.thisDefs.get(name));      // should not fail
+      }
+
+      // TODO: will we support hot compilation? or we can just assume this is a CompiledVar
```
ice1000: Likely not

